PR #28, opened 2026-05-09
test: smoke daemon ritual flows

BunsDev:
## Summary
- add real CLI smoke coverage for daemon restart before replaying persisted session output
- add `coven daemon restart` as the user-facing restart path
- update README/docs/roadmap to reflect restart and shipped smoke coverage

## Verification
- watched the smoke test fail first because `coven daemon restart` did not exist
- `cargo fmt --check`
- `cargo test --workspace --locked` — 111 unit + 1 smoke passed
- `npx --yes -p vitest@4.1.5 vitest run src/*.test.ts` — 84 passed
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ Prefer copy/paste commands?
 cd /path/to/your/project
 coven doctor
 coven daemon start
+coven daemon restart
 coven run codex "fix the failing tests"
 coven run claude "polish this UI"
 coven sessions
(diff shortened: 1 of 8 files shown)


BunsDev:
Reviewed Copilot feedback and applied the security fix in `c7d41b5`: daemon stop/restart now verifies the status socket reports the recorded pid before signaling, refuses unverified live PIDs, waits for the old daemon to exit before clearing status/socket, and only clears stale metadata when the recorded PID is gone. Verified locally with Rust gates, plugin Vitest, secret scan, and `git diff --check`; GitHub CI/CodeQL/Secret Guard are green.

PR #113, opened 2026-05-21
feat(chat): syntax-highlight fenced rust + ts/js code blocks

BunsDev:
## Summary
- Tokenize fenced code-block content when the opening fence tags a known language (`rust`/`rs`, `js`/`jsx`/`ts`/`tsx`/`javascript`/`typescript`); map keyword/string/number/comment/attribute roles onto five new brand-aligned `SYNTAX_*` tokens in `theme.rs`.
- Thread block-comment state across lines so `/* … */` stays styled the whole way; state resets at every opening fence so adjacent blocks never bleed.
- Hand-rolled, zero-dep tokenizers (~580 LOC incl. tests) — no new dependencies, no binary-size hit; `ACCENT_BLUE` is no longer `#[allow(dead_code)]`.
- Unknown / missing language tags fall back to the existing verbatim render, so the change is purely additive.

## Test plan
- [x] `cargo test -p coven-cli` — 438 tests pass, 1 ignored (the preview).
- [x] `cargo clippy -p coven-cli --all-targets` — no new warnings.
- [x] Visual smoke via `cargo test -p coven-cli preview_chat_highlight_to_truecolor_stdout -- --ignored --nocapture` — TrueColor preview confirms keyword/string/number/comment/attribute styling and multi-line block comments span correctly.
- [ ] Try `coven chat` against a real session; paste a fenced ```` ```rust ```` block from an agent reply.

### DIFF
--- a/crates/coven-cli/src/theme.rs
+++ b/crates/coven-cli/src/theme.rs
@@ -30,9 +30,6 @@ pub mod brand {
         g: 0xBD,
         b: 0xED,
     };
-    // Defined for future use; consumed via the `DANGER` / `SUCCESS` semantic tokens
-    // in later phases (destructive prompts, ready-state indicators).
-    #[allow(dead_code)]
     pub const ACCENT_BLUE: Rgb = Rgb {
         r: 0x0A,
         g: 0x84,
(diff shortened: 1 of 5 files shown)


BunsDev:
Pushed [82aff39](https://github.com/OpenCoven/coven/commit/82aff39) extending coverage to Python (`python`/`py`) and Bash (`bash`/`sh`/`shell`/`zsh`).

**Python:** keywords, `#` comments, prefixed strings (`f"…"`, `rb"…"`), decorators, complex literals (`1j`), and multi-line triple-quoted strings — the latter required extending `TokenizerState` with `in_python_triple: Option<TripleKind>` so closers have to match the opener's flavor.

**Bash:** keywords, shebang/`#` comments, single-quoted strings (fully literal per bash semantics) + double-quoted, and `$VAR` / `${VAR}` / `$?`-style variable references styled as Attribute.

454 unit tests pass (was 438), 1 ignored (the TrueColor preview, now exercising all four languages).

Issue #142, opened 2026-05-28
Daemon: fail closed on symlinked COVEN_HOME and non-socket socket path (AUTH.md hardening gap)


maplesyzzurp:
## Summary

`docs/AUTH.md` → "Current hardening gap" lists the fail-closed checks the Rust daemon should perform before creating, binding, or removing daemon state. Two are still unimplemented in `crates/coven-cli/src/daemon.rs`:

- **A `COVEN_HOME` that resolves through a symlink is not rejected.** `ensure_private_coven_home` calls `create_dir_all` + `set_permissions(0700)` and will follow a symlinked home, chmod-ing and populating the link target.
- **An existing socket path that is a symlink or non-socket file is not rejected.** `bind_api_socket` does `if socket_path.exists() { remove_file(socket_path) }` — it blindly removes whatever sits at `coven.sock` before binding, with no symlink/socket-type guard.

The permission hardening from the same list (`0700` home, `0600` socket/status) and the body-cap / loopback-TCP work already landed — this is only the residual symlink/type guards. The `COVEN_HOME` *ownership* (euid) check is intentionally out of scope here since it needs a `libc::geteuid` dependency.

## Why it matters

With a `0700` home these are defense-in-depth rather than a same-user break, but they are exactly the conditions AUTH.md says the daemon should "fail closed" on before broad distribution: a planted symlink at the home or socket path could redirect daemon state (socket, status, SQLite ledger) outside the trusted directory, and the unconditional `remove_file` will delete an unrelated regular file that happens to occupy the socket path.

## Proposed fix (implemented, std-only)

- `ensure_private_coven_home`: `symlink_metadata` the home; bail if it is a symlink.
- `bind_api_socket`: `symlink_metadata` the socket path; bail on a symlink or any non-socket file; only `remove_file` a genuine stale socket; treat `NotFound` as fresh.

No new dependencies, no `unsafe`, Unix-gated. Three new `#[cfg(unix)]` tests cover a symlinked home, a symlinked socket path (and assert the link target is not deleted), and a non-socket file at the socket path.

Verification (Linux): `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace --locked` all green (696 unit + 4 smoke; the 3 new tests pass, no regressions).

## Branch

A branch is ready to PR when the contribution window reopens (July 2026, per CONTRIBUTING):
[...]

PR #144, opened 2026-05-29
Harden daemon: fail closed on symlinked COVEN_HOME and non-socket socket path

maplesyzzurp:
## Summary

Closes the two still-open `docs/AUTH.md` "Current hardening gap" fail-closed checks in `crates/coven-cli/src/daemon.rs`:

- `ensure_private_coven_home` now refuses a `COVEN_HOME` that is a symlink, so following it cannot redirect daemon state (socket, status, SQLite ledger) outside the trusted directory.
- `bind_api_socket` now inspects the socket path with `symlink_metadata` and refuses a symlink or any non-socket file, instead of calling `remove_file` on whatever happens to be there. Only a genuine stale socket is replaced.

## Why

`docs/AUTH.md` lists these exact conditions under "Before broad distribution, Rust should fail closed when ... `COVEN_HOME` resolves through a symlink ... an existing socket path is a symlink or non-socket file." The TypeScript plugin already validates the socket trust anchor; this brings the Rust daemon in line. The euid-ownership check from the same list is intentionally left out of this PR because it needs a `libc::geteuid` dependency.

## Changes

- `ensure_private_coven_home`: `symlink_metadata` guard; bail when the home is a symlink.
- `bind_api_socket`: `symlink_metadata` + `FileTypeExt::is_socket`; bail on a symlink or non-socket; only `remove_file` a genuine stale socket; treat `NotFound` as fresh.
- 3 new `#[cfg(unix)]` tests (symlinked home, symlinked socket path with target-not-deleted assertion, non-socket file at the socket path).

std-only — no new dependencies, no `unsafe`, Unix-gated.

## Verification

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace --locked` all green on Linux: 696 unit + 4 smoke tests pass, including the 3 new ones, no regressions.

Tracked by #142.

---

Note: I see PRs are frozen until July per CONTRIBUTING — opening this for visibility since it is a small, isolated security hardening that closes a documented gap. Completely fine to close or hold it until the window reopens; the branch will keep.

### DIFF
--- a/crates/coven-cli/src/daemon.rs
+++ b/crates/coven-cli/src/daemon.rs
@@ -10,7 +10,7 @@ use std::time::{Duration, Instant};
 use std::io::{BufRead, BufReader, Read};
 #[cfg(unix)]
 use std::os::unix::{
-    fs::PermissionsExt,
+    fs::{FileTypeExt, PermissionsExt},
     net::{UnixListener, UnixStream},
 };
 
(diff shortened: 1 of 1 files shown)


maplesyzzurp:
/Lock-in 
Shoutout to C#aude 4.8 M8

PR #184, opened 2026-06-06
[codex] add Coven task manager skill

BunsDev:
## What changed

- Adds the canonical `coven-task-manager` skill with metadata and default paused Codex automation templates.
- Adds `task-manager.mjs` to generate task freshness reports, install local skill symlinks, and install default automation TOMLs.
- Adds a Thread Coordination section that groups simultaneous work by shared session, familiar, repo, and branch so agents reconcile existing lanes before spawning more work.
- Updates the Cave skill scanner to follow symlinked local skill directories.

## Why

Coven task hygiene needed to handle multiple active agent threads intelligently. The report now surfaces collisions directly, and the skill tells agents to build a thread ledger, resume viable existing sessions, park duplicate lanes, and avoid parallel writes to the same repo/branch.

## Validation

- `node skills/coven-task-manager/task-manager.test.mjs`
- `cargo test -p coven-cli scan_skills_follows_symlinked_skill_dirs`

### DIFF
--- a/crates/coven-cli/src/cockpit_sources.rs
+++ b/crates/coven-cli/src/cockpit_sources.rs
@@ -263,9 +263,12 @@ pub fn scan_skills(coven_home: &Path) -> Result<Vec<SkillDto>> {
         match fs::metadata(&dir) {
             Ok(meta) if meta.is_dir() => {}
             Ok(_) => continue,
+            Err(err) if err.kind() == io::ErrorKind::NotFound => continue,
             Err(err) => {
-                return Err(err).with_context(|| format!("failed to inspect {}", dir.display()));
+                return Err(err)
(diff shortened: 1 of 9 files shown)


BunsDev:
Reviewed after rebasing onto current main and applying hardening fixes. The diff is now focused on task-manager skill support plus symlinked skill scanning. I removed the pinned/guessed automation model, added status validation, and changed symlink replacement to unlink rather than recursive removal. Local verification: git diff --check, node --test skills/coven-task-manager/task-manager.test.mjs, cargo fmt --check, focused scan_skills tests, and cargo clippy -p coven-cli --all-targets -- -D warnings. GitHub checks are green. No unresolved review threads. GitHub would not accept an approval from this authenticated account because it is the PR author's account.

PR #271, opened 2026-06-28
feat(adapter): add `coven adapter login codex` with OAuth port preflight

BunsDev:
Fixes the recurring "Failed to bind port 1455" error you hit whenever a `codex login` flow gets killed mid-OAuth and leaks an orphan listener on the callback port.

## The problem

`codex login` opens a local OAuth callback server on **port 1455**. If the flow gets killed mid-auth (esc, SIGINT, browser closed, network blip) the listener sometimes leaks as an orphan process. The next `codex login` attempt then crashes with:

```
Codex OAuth failed: Failed to bind port 1455: Address already in use (os error 98)
```

(`os error 48` on macOS, `os error 98` on Linux — same condition.)

Upstream Codex has no auto-recover or fallback-port logic, so until that lands in `@openai/codex` itself, this shim handles it from our side.

## The fix

New subcommand: `coven adapter login <adapter>`. Today the only adapter that needs preflight is `codex`. Behavior:

1. Try to bind `127.0.0.1:1455` ourselves
2. If `AddrInUse`, identify the holding pid via `lsof -ti tcp:1455`
3. Use `sysinfo` to fetch the holder's process descriptor (name + argv)
4. **Only kill it if its argv/path clearly contains the token `codex`** — split on whitespace and path separators, so `vscodex`, `not-codex-tool`, etc. don't match
5. SIGTERM (1s grace), then SIGKILL if still alive
6. Wait up to 1.5s for the port to free up
7. `exec codex login`

If we can't ID the holder (no lsof, ambiguous results), or if it's clearly NOT codex, we **refuse to kill** and print a clear instruction:

```
$ coven adapter login codex
Coven codex login
  Preflight: checking port 1455... held
Error: port 1455 is held by pid 83156 (`Python`), which does NOT look like a codex OAuth helper. Refusing to kill it. Run `lsof -i tcp:1455` to investigate, then either stop that process or rerun `coven adapter login codex` once it is gone.
```

## Why this shape

- **Not `coven doctor`** — users hit this error mid-frustration trying to log in. They don't want "go run a separate doctor command first". One command that just works.
- **Not changing Codex's port** — it's hard-coded upstream; we can't.
- **`coven adapter login codex` not `coven codex login`** — reuses the existing `coven adapter ...` namespace (`list`/`doctor`/`install`/`login`).
- **Conservative kill heuristic** — substring-matching `codex` would have false positives (`vscodex`, `@openai/codex-darwin-arm64`). Token-level matching is bulletproof and explicit.

## Tests

New module `crates/coven-cli/src/codex_login.rs` with **7 unit tests**, all passing:

- `port_is_free` correctly identifies free vs held ports (uses ephemeral port to avoid flakes)
- `descriptor_looks_like_codex` recognizes `codex`, `codex.js`, `@openai/codex`
- `descriptor_looks_like_codex` rejects `vscodex`, `not-codex-tool`, empty strings
- `descriptor_looks_like_codex` handles path separators (`/usr/local/bin/codex`)
- `pid_holding_port` returns None when port is free (also covers lsof-missing case)
- `preflight_codex_oauth_port` returns `PortFree` when 1455 is unbound (auto-skips if 1455 is occupied — no flake)

Plus full crate test suite: **856 unit + 4 + 11 = 871 tests, all green.** No regressions.

## Manual verification

- ✅ `coven adapter login --help` shows the new subcommand
- ✅ Preflight with port FREE: reports "free" and exec's codex
- ✅ Preflight with port HELD by non-codex (python): refuses to kill, prints lsof instructions
- ✅ `cargo clippy -- -D warnings` clean
- ✅ `cargo fmt --check` clean

## What it looks like in practice

```
$ coven adapter login codex
Coven codex login
  Preflight: checking port 1455... freed (killed stale codex pid 12345)
  Launching `codex login`...

  > Sign in via browser or paste your existing token...
```

## Not in scope (separate follow-ups)

- A shim for `claude login` — Anthropic CLI doesn't have this failure mode (uses bearer-token paste, no local callback server)
- Surfacing the preflight in cave's UI — this PR ships the CLI handler; cave can wrap it later if useful

Once this lands, you can rebuild + install with the usual `cargo install --path crates/coven-cli` or wait for the next coven release stamp.

### DIFF
--- a/crates/coven-cli/src/codex_login.rs
+++ b/crates/coven-cli/src/codex_login.rs
@@ -1,0 +1,325 @@
+//! `coven adapter login <adapter>` shim.
+//!
+//! Wraps a vendor CLI's own login flow with our preflight cleanup. Today the
+//! only adapter that needs this is Codex (OpenAI), whose `codex login` flow
+//! spins up a local OAuth callback server on port **1455** and fails hard
+//! (`EADDRINUSE`) when a previous `codex login` was killed mid-flow and the
+//! orphan listener is still around.
+//!
+//! Preflight behavior:
+//! 1. Try to bind 127.0.0.1:1455 ourselves. If that succeeds, the port is
+//!    free — drop the listener and proceed straight to `codex login`.
+//! 2. If bind fails with `AddrInUse`, identify the holding process via `lsof`
+//!    and only kill it if its command line clearly identifies it as a `codex`
+//!    OAuth helper. Never kill an unidentified process — we'd rather print a
+//!    clear error than terminate something unrelated that happens to be on
+//!    the same port.
+//! 3. Wait for the port to actually free up (poll up to ~1.5s), then exec
+//!    `codex login`.
+//!
+//! This shim is deliberately conservative: any time we're not 100% sure the
+//! port-holder is a codex OAuth listener, we bail with instructions for the
+//! user.
+
+use anyhow::{anyhow, Context, Result};
+use std::io::ErrorKind;
+use std::net::TcpListener;
+use std::process::Command;
+use std::time::{Duration, Instant};
+
+/// The fixed port that Codex CLI's OAuth callback server tries to bind.
+///
+/// This is hard-coded in the upstream `@openai/codex` package. There is no
+/// configuration to make Codex pick a free port instead, which is why we need
+/// this shim in the first place.
+pub const CODEX_OAUTH_PORT: u16 = 1455;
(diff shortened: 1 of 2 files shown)


BunsDev:
Closing — moving the fix to coven-cave's UI instead of the CLI per Val's direction. Will reopen here later if the CLI shim is also wanted.

PR #297, opened 2026-07-06
fix: CLI UX polish — front-door guards, session-ref UX, output discipline, honest help

BunsDev:
## Summary

Result of a comprehensive UI/UX review of the coven CLI (help surface, first-run/onboarding, error paths, TUI live surfaces, output/exit-code discipline — each audited hands-on against a scratch `HOME` plus code review). This PR lands the high-value, low-risk fixes; the remaining findings are listed at the bottom as follow-up candidates.

## Worst failure mode fixed (P0)

The top-level `[PROMPT]...` catch-all swallowed everything clap didn't recognize, so with stdin at `/dev/null`:

- `coven sesions` → **launched a real Codex session** (typo became an AI task; Cast's "press enter to cast" gate fails open on EOF), left an orphaned `running` session behind, and auto-spawned a daemon.
- `coven --json` → same: unknown flags became prompts.
- `coven run` exited **0** even when the harness session failed — `coven run ... && deploy` in CI proceeded past failures.

Now: typos get a did-you-mean error, stray flags error, Cast refuses to run without an interactive terminal, and `coven run` exits with the harness exit code (stream-json result events are unchanged; a one-line `session failed (exit code N)` goes to stderr on the human path).

## Other fixes

**Session rituals** — `attach`/`summon`/`archive`/`sacrifice` accept unique id prefixes (`coven attach 9099`); `sacrifice` validates the id *before* demanding `--yes`; `archive` is idempotent; the "still running" refusals now explain the `coven daemon restart` orphan-recovery remedy (previously a dead end: the error said "stop it" but no stop verb exists); `attach` says whether it's following or replaying and how to detach/resume.

**Scripting** — `sessions` plain-table hints move to stderr so stdout stays parseable; `RITUAL` column renamed `STATE` (the value was `active`/`archived`, not a ritual); signal-killed coven-code delegation exits 1; `wt`/`claim`/`hooks` outside a git repo print a one-line remedy instead of raw `git failed: fatal:` output.

[...]

Issue #309, opened 2026-07-06
No shell completions (`clap_complete`) and no `CLICOLOR_FORCE`/`--color` override


BunsDev:
From the #297 CLI UX audit (follow-up candidate 6). Two related CLI-conventions gaps:

## 1. Shell completions

No completion generation is wired up. `clap_complete` can derive bash/zsh/fish completions from the existing clap definitions — typically exposed as `coven completions <shell>` plus an install note in the docs.

## 2. Color control

Color output honors auto-detection only. There is no `--color=always|never|auto` flag and `CLICOLOR_FORCE`/`NO_COLOR` conventions aren't fully respected, so piped output can't opt back into color (e.g. through a pager) and CI can't force it off deterministically.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #312, opened 2026-07-06
fix: keep --stream-json stdout JSONL-only on non-claude harnesses

### LINKED ISSUE #307: `--stream-json` on non-claude harnesses interleaves raw PTY bytes with JSONL on stdout
From the #297 CLI UX audit (follow-up candidate 1).

## Problem

With a claude harness, `--stream-json` emits clean JSONL on stdout. With non-claude harnesses (codex, etc.), raw PTY bytes are interleaved with the JSONL on the same stdout stream — while the docs declare the stream format stable. Any consumer parsing line-delimited JSON gets corrupted input depending on which harness happens to run.

## Expected

stdout carries only the stable JSONL stream regardless of harness. PTY output needs captured-PTY routing (kept off stdout, or wrapped into stream events) so the contract holds for every adapter.

## Pointers

- Follow-up list in #297; claude one-shot stream-json path lives in `pty_runner`.

BunsDev:
Closes #307 (from the #297 UX audit, follow-up 1).

## Problem

With non-claude harnesses, `coven run --stream-json` ran the harness attached to the PTY: raw harness bytes hit stdout interleaved with the JSONL frames, corrupting the stream the docs declare stable. Consumers parsing line-delimited JSON got garbage whenever the harness wasn't claude.

## Fix

- **Stream mode always launches the one-shot non-interactive entrypoint** — TTY detection no longer applies; an interactive TUI behind a JSONL pipe was never meaningful.
- **The harness runs fully captured** (`pty_runner::run_captured`, no PTY): stdout is collected, never forwarded. No PTY also means no ANSI echo (non-interactive entrypoints already disable color). Signal-killed children map to exit 1, matching the existing delegation rule.
- **Captured output becomes a single `assistant` frame** before the closing `result`, so consumers now get the harness text *in* the stream (previously documented as unavailable). Additive per STREAM-JSON.md's stability policy; the doc's Harness-behavior section is updated.
- **Harness stderr forwards to coven's stderr** — visible, but out of band of the stream contract.

The human (non-stream) path is untouched: still attached PTY.

## Testing

- New smoke test drives a fake codex through `coven run codex --stream-json` and asserts **every stdout line parses as JSON** with the init → user → assistant → result sequence and the harness output inside the assistant frame. This fails against the old attached-PTY behavior.
- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace --locked`, `scripts/check-secrets.py` — all green.

## Drive-by test-determinism fix

`adapter_install_hermes_writes_trusted_manifest` (from #306) diagnosed hermes against the real PATH, so its expected exit-1 flipped on machines that actually have a `hermes` binary (observed locally). It now diagnoses against an empty PATH.

### DIFF
--- a/crates/coven-cli/src/main.rs
+++ b/crates/coven-cli/src/main.rs
@@ -1978,16 +1978,42 @@ fn run_session(
         .as_ref()
         .filter(|s| s.system_prompt_flag.is_some())
         .and(familiar_ctx.as_ref());
+    // Stream mode's stdout contract is JSONL-only, so it always launches the
+    // one-shot non-interactive entrypoint regardless of TTY detection — an
+    // interactive TUI behind a JSONL pipe would be garbage either way.
+    let launch_mode = if stream_json {
+        harness::HarnessLaunchMode::NonInteractive
+    } else {
+        harness_launch_mode_for_stdio()
+    };
     let command = pty_runner::build_harness_command_with_conversation(
         &selected_harness.id,
         &effective_prompt,
         &cwd,
-        harness_launch_mode_for_stdio(),
+        launch_mode,
         conversation_hint.as_ref(),
         familiar_for_args,
         launch_options,
     )?;
-    match pty_runner::run_attached(&command) {
+    // Stream mode captures the harness instead of attaching it to the PTY:
+    // raw harness bytes on stdout would interleave with the JSONL frames and
+    // corrupt the stream for every consumer (#307). The captured text is
+    // re-emitted below as an `assistant` frame.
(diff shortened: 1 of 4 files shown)


BunsDev:
Superseded by #315, which merged the same fix for #307 (`--stream-json` stdout kept JSONL-only on non-claude harnesses via an additive `output` event). Closing this duplicate — it was developed in parallel by a concurrent session on the same checkout. The completions/color work (#309) from that same session is still valuable and not duplicated.

Issue #311, opened 2026-07-06
Output polish: `^D` artifact on piped `coven run`; patch cancellation surfaces as `Error:`; key=value dev-log voice on human status lines


BunsDev:
From the #297 CLI UX audit (follow-up candidate 8). Three small output-discipline nits, bundled:

1. **`^D` artifact echoed on piped `coven run`** — the EOF control-char echo from the PTY leaks into visible output when stdin is a pipe.
2. **`patch` cancellation surfaces as an `Error:`** — a user-initiated cancel is reported in failure voice; it should read as a neutral "cancelled" outcome (exit code can stay nonzero for scripts).
3. **key=value dev-log voice on some human status lines** — a few human-path status lines still print `key=value` telemetry style instead of prose, inconsistent with the output discipline #297 established elsewhere.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

BunsDev:
Reopening: PR #319 landed items 1 (`^D` artifact) and 2 (neutral patch-cancel voice) but not item 3 (the `key=value` human status lines). #319 merged before the item-3 completion could be pushed, so this issue auto-closed with item 3 still on main. A follow-up PR converting the daemon-status human lines to prose is incoming and will close this properly.

PR #321, opened 2026-07-07
fix(tui): legacy chat paper cuts — Ctrl+C draft, help clipping, /attach prefixes

### LINKED ISSUE #310: Legacy chat TUI: Ctrl+C kills running session when clearing a draft; help overlay clips Keys; `/sessions` shows ids `/attach` cannot use
From the #297 CLI UX audit (follow-up candidate 7). Lower priority: the legacy TUI is behind `COVEN_LEGACY_TUI` now that coven-code is the flagship interactive UI — but while the flag exists these are real paper cuts:

1. **Ctrl+C kills the running session even when its intent was clearing a draft.** With text in the input box, Ctrl+C should clear the draft (or require a second press to kill), not tear down live work.
2. **The help overlay clips its Keys section** — the bindings list is cut off and can't be scrolled into view.
3. **`/sessions` overlay shows truncated session ids that `/attach` can't consume.** Either display attachable prefixes (the CLI rituals accept unique prefixes as of #297) or make `/attach` resolve the truncated form.

BunsDev:
Closes #310 (from the #297 UX audit, follow-up 7). Lower priority — the legacy chat TUI is behind `COVEN_LEGACY_TUI` now that coven-code is the flagship UI — but while the flag exists these are real paper cuts.

## 1. Ctrl+C kills live work while clearing a draft

`handle_interrupt` unconditionally killed the active session on the first press *and* cleared the draft. Now it escalates in priority order — **pending Cast confirmation → draft input → active session** — and stops after the first thing it clears. A second press within the rearm window still exits. So Ctrl+C with a half-typed message discards the draft and leaves the running session alone.

## 2. Help overlay clips its Keys section

The overlay used a fixed height cap shorter than its content, so the bottom bindings (`Ctrl+L`, `Ctrl+D`) were cut off with no way to scroll. It's now sized to its content height (+borders), and widened so the longest `command  description` row no longer wraps — wrapping is what had made a naive line-count height undercount the real rows.

## 3. `/sessions` ids can't be `/attach`ed

The overlay showed an ellipsis-truncated id (`9099a1b2-3c…`) that `/attach` couldn't consume. Now it shows a clean 8-char prefix, and `/attach` resolves a unique prefix to the full session id (ambiguous prefixes are reported), mirroring the CLI ritual verbs' prefix handling.

## Testing

- Ctrl+C with a draft clears it and sends no kill (session stays active); with no draft, the press interrupts the session.
- Help overlay renders its full Keys section (`Ctrl+L`, `Ctrl+D`, `Exit Coven chat`) on a tall terminal.
- `/attach` resolves a unique prefix to the full id; an ambiguous prefix is reported without attaching; unit test for the id-prefix helper.
- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test -p coven-cli` (945 unit + 19 smoke), `scripts/check-secrets.py` — all green.

Rebased on current main (through #311 and the RUSTSEC-2026-0204 crossbeam bump).

### DIFF
--- a/crates/coven-cli/src/tui/chat/app.rs
+++ b/crates/coven-cli/src/tui/chat/app.rs
@@ -900,19 +900,21 @@ impl App {
             return InterruptOutcome::Quit;
         }
 
-        // First press: cancel everything cancellable, then arm exit.
-        let had_pending = self.cancel_pending_cast_confirmation();
-        let had_input = !self.input.is_empty();
-        let interrupted_session = self.interrupt_active_session();
-        self.input.clear();
-        self.cursor_pos = 0;
-        self.slash_suggestion_index = 0;
-        self.slash_popup_dismissed = false;
-
-        let advisory = if interrupted_session {
-            "Interrupt sent. Press Ctrl+C again to exit."
-        } else if had_pending || had_input {
+        // First press: clear the most transient state and stop there, so a
+        // Ctrl+C meant to discard a draft or a Cast prompt never tears down
+        // live work. Only when there is nothing composed to clear does the
+        // first press interrupt the running session. Priority: pending Cast
+        // confirmation → draft input → active session.
+        let advisory = if self.cancel_pending_cast_confirmation() {
             "Cleared. Press Ctrl+C again to exit."
+        } else if !self.input.is_empty() {
+            self.input.clear();
+            self.cursor_pos = 0;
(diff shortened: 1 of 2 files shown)


BunsDev:
Closing in favor of #320 (opened first, equivalent coverage of #310). Both were produced by parallel sessions.

PR #323, opened 2026-07-07
fix: output polish — no `^D` artifact on piped runs, neutral patch cancel, prose status lines

### LINKED ISSUE #311: Output polish: `^D` artifact on piped `coven run`; patch cancellation surfaces as `Error:`; key=value dev-log voice on human status lines

BunsDev:
## Summary

Closes #311 (follow-up candidate 8 from the #297 CLI UX audit).

### 1. `^D` artifact on piped `coven run`
`portable-pty`'s master writer writes `\n` + VEOF on drop so the harness sees EOF — but with the PTY line discipline still in echo mode, that byte pair (and the piped input itself) was echoed back into the visible output as a stray `^D`. Attached runs now clear `ECHO` on the PTY when stdin is not a terminal. No-op on Windows (ConPTY has no termios echo) and when stdin is a real TTY. Regression test spawns `cat` under a real PTY and asserts no EOT byte and no double-echoed input.

### 2. `patch` cancellation surfaced as `Error:`
Declining either `coven patch` confirmation prompt used `anyhow::bail!`, which the top level printed as `Error: cancelled before harness launch`. A user-initiated cancel now prints a neutral `Cancelled. Nothing was launched.` — exit code stays `1` so scripts still see it didn't run.

### 3. key=value dev-log voice on human status lines
- `coven doctor`'s Daemon section now matches the rest of doctor's `[OK]`/`[!!]` prose (`[OK] running — pid 123, socket …`, stale gets a `coven daemon restart` remedy line).
- `coven logs prune` speaks prose (`Pruned 3 raw artifacts (older than …) and 12 events …`), with correct singular/plural.
- **Intentionally unchanged:** `coven daemon status` key=value output — docs describe it as the scriptable status surface (docs/daemon/health.md, cli-doctor.md), `scripts/test-cli-prepublish.mjs` and smoke tests parse it, and follow-up #297-candidate 5 (`--json` for `daemon status`) is the right place to migrate scripts before changing it.

## Testing
- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace --locked`, `python scripts/check-secrets.py` — all green.
- New tests: `disable_pty_echo_suppresses_input_and_eot_echo` (real PTY), `count_noun_pluralizes_for_human_status_lines`; updated the doctor smoke assertion to the new prose.

### DIFF
--- a/crates/coven-cli/src/main.rs
+++ b/crates/coven-cli/src/main.rs
@@ -922,19 +922,27 @@ fn run_doctor() -> Result<()> {
     match daemon::background_server_status(&home)? {
         Some(daemon::DaemonStatusState::Running(status)) => {
             let health = api::health_response(Some(status.clone()));
-            println!(
-                "  status=running ok={} pid={} socket={}",
-                health.ok, status.pid, status.socket
-            );
+            if health.ok {
+                println!(
+                    "  [OK] running — pid {}, socket {}",
+                    status.pid, status.socket
+                );
+            } else {
+                println!(
+                    "  [!!] running but failing health checks — pid {}, socket {}",
+                    status.pid, status.socket
+                );
+            }
         }
         Some(daemon::DaemonStatusState::Stale(status)) => {
             healthy = false;
             println!(
-                "  status=stale ok=false pid={} socket={}",
+                "  [!!] stale — pid {} no longer looks healthy (socket {})",
                 status.pid, status.socket
             );
+            println!("       Recover with `coven daemon restart`.");
         }
-        None => println!("  status=stopped"),
+        None => println!("  [OK] stopped — optional; start with `coven daemon start`"),
     }
 
     let repos_config = repos_config::load_with_settings(&home, settings::cached())?;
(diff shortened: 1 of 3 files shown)


BunsDev:
Superseded — all three #311 items are already on main: items 1 (`^D` artifact) and 2 (neutral patch-cancel) landed via #319, and item 3 (prose daemon-status lines) via #324. #311 is closed as completed. Closing this duplicate, developed in parallel by a concurrent session on the same checkout; no work is lost (the branch remains). Leaving your unrelated store-vacuum PR #322 untouched.

Issue #329, opened 2026-07-07
[Windows] Every PTY-run harness hangs at startup: PTY runner never answers VT queries (ESC[6n); sessions stay "running" forever with zero events


mithril-logic:
## Environment

- Windows 11 Pro (10.0.26200)
- @opencoven/cli 0.0.53 (npm, registry latest), daemon run with `RUST_LOG=debug`
- Harnesses: codex CLI and Claude Code, both current, both work normally when run standalone in a real terminal

## Symptom

`coven run` (any harness — verified with codex and claude) starts a session that produces **zero events, forever**. The harness process spawns and sits there; the ledger session stays `running` indefinitely; `coven attach` shows nothing. No timeout, no error, no output.

## Repro

1. Familiar bound to codex (`harness = "codex"`, any model) in `~/.coven/familiars.toml`.
2. `coven run <familiar> "ping"` (a prompt that normally returns "pong" in seconds).
3. Query the ledger (`~/.coven/coven.sqlite3`): the session row exists, status `running`, **0 rows of events**, hours later.
4. Same result with a claude-backed familiar. Both harnesses respond within seconds when invoked directly in Windows Terminal.

## Root cause (smoking gun)

Capturing the harness's raw output at spawn shows it emits its terminal init sequence and then blocks:

```
ESC[?9001h ESC[?1004h ESC[6n
```

That trailing `ESC[6n` is a **Device Status Report query** — the TUI (crossterm-based, both codex and claude use it) is asking "where is the cursor?" and *waits for the terminal to reply* `ESC[<row>;<col>R` on stdin. A real console (or a full ConPTY client that pumps responses) answers; coven's PTY runner never writes a reply, so the harness blocks in its first paint before ever reading the prompt. Hence: process alive, zero events, session `running` forever.

## Suggested fixes

1. **Answer VT queries in the PTY reader**: on seeing `ESC[6n` (and friends — CPR, DA `ESC[c`), write a plausible reply to the child's stdin. Small, fixes every crossterm harness at once.
2. **Prefer headless modes on Windows** (or everywhere): the piped `claude -p --input-format stream-json --output-format stream-json` path already exists in `crates/coven-cli/src/pty_runner.rs` (~:283, "spawn claude in stream-json mode") — routing Windows through headless/piped modes sidesteps ConPTY entirely and is more robust for a daemon anyway.

## Secondary findings from the same debugging session

[...]

Issue #195, opened 2026-06-08
fix: body cap + I/O timeouts on Unix socket and Windows pipe transports


BunsDev:
## Context
Only the TCP transport currently sets `MAX_TCP_BODY_BYTES` / `TCP_IO_TIMEOUT`. Unix socket and Windows pipe transports have no equivalent caps.

## Task
Apply consistent body cap and I/O timeout limits to all transports.

## References
- Tracking: #147 (Tier 3, item 10)

PR #336, opened 2026-07-11
Fix Windows Cave and Codex session reliability

CompleteDotTech:
## What changed

- use `codex exec` for prompted Windows CLI runs
- add a real Windows named-pipe transport for Cave/chat instead of the Unix-only error branch
- parse HTTP responses by `Content-Length` rather than waiting for pipe EOF
- bound named-pipe connect and response waits so doctor/status/Cave cannot wait forever
- serve accepted Windows pipe connections on independent handler threads so one stalled client cannot starve polling
- claim the Windows pipe before writing daemon metadata or orphaning persisted sessions
- run Windows non-interactive daemon children with ordinary pipes instead of ConPTY
- preserve stdout/stderr observation, exit recording, and Windows Job Object cleanup
- normalize Windows extended-length project paths before exposing them to Cave
- honor PATHEXT across PATH directories so codex.exe wins over an earlier npm codex.CMD shim
- route the Windows Codex labeled transcript onto the Cave-consumed stdout channel without duplicating the final answer
- add regression coverage for framing, timeouts, repeated health requests, Windows Cave transport, launch mode, output persistence, terminal state, polling backoff, and rendering

## Root cause

Several Windows-specific gaps combined into the apparent Cave hang:

1. Cave's daemon client was compiled to return `Coven daemon chat is only implemented on Unix-like systems` on Windows.
2. Doctor/status read named-pipe HTTP responses with `read_to_string()`, waiting for EOF instead of the declared body length.
3. Pipe connection establishment was unbounded and the Windows daemon handled accepted clients serially.
4. A replacement daemon could mark live rows orphaned before proving it owned the pipe.
5. One-shot codex exec launched through ConPTY could terminate as u32::MAX with no captured output; direct piped execution works.
6. Windows PATH resolution selected an earlier npm codex.CMD shim even when PATHEXT preferred a later native .EXE, causing invalid batch argument failures.
7. Codex emitted its labeled transcript on stderr and a duplicate plain answer on stdout; Cave only recognized labeled assistant text on Coven's stdout channel, so successful runs were misreported as empty output.

## Impact

Cave can now launch non-interactive Codex work through the Windows daemon, poll persisted output events, observe terminal state, and recover from transient pipe failures. Doctor/status probes are framed and bounded. Claude and non-Windows launch behavior remain unchanged.

## Validation

- focused Windows/Cave reliability tests: all passed
- full `cargo test -p coven-cli --quiet`: 825 passed, 11 failed, 1 ignored
- the 11 failures are pre-existing Windows assertions that expect literal `codex`/`claude` executable names while this machine resolves absolute `.EXE`/`.CMD` paths
- direct `codex exec hi` on the affected machine returns a response and exits successfully
- isolated daemon/API diagnostics reproduced both the original ConPTY failure and pipe starvation, which the new piped-launch and concurrent-handler paths address
- split-stream smoke after 7979630: exit 0, exactly one CAVE_FINAL_OK assistant line on stdout, zero stderr bytes, and a successful result event
- installed Cave session 470e0215-83fe-45d4-9f1d-7e065d863341: response persisted after 15,003ms with isError: false and no leftover Coven/Codex child

### Final installed-Cave verification

Cave was relaunched with `COVEN_BIN` pointing at the PR build. That exposed one final foreground CLI ConPTY path, now fixed in `8ba90c6`. The exact Cave-style command (`coven run codex --stream-json --model openai/gpt-5.5 ...`) then emitted init/user events, returned `CAVE_PIPE_OK`, emitted a successful result, and exited 0 in 17 seconds.

## Related work

- Partially addresses #329 (Windows PTY startup hang, zero-event running sessions, and extended-path leakage). The issue also covers Claude and should remain open until that path is independently verified.
- Related stream contract: #307 / #315.
- Related Windows/local transport hardening: #195 / #207, #242, and #243.

### DIFF
--- a/crates/coven-cli/src/api.rs
+++ b/crates/coven-cli/src/api.rs
@@ -3341,11 +3341,15 @@ mod tests {
         assert_eq!(runtime.launches.borrow()[0].prompt, "hello coven");
         assert_eq!(
             runtime.launches.borrow()[0].project_root,
-            project_root.canonicalize()?.to_string_lossy()
+            project::canonical_project_root(&project_root)?.to_string_lossy()
         );
         assert_eq!(
             runtime.launches.borrow()[0].cwd,
-            cwd.canonicalize()?.to_string_lossy()
+            project::resolve_inside_root(
+                &project::canonical_project_root(&project_root)?,
+                Some(&cwd)
+            )?
+            .to_string_lossy()
         );
         assert!(list.body.contains(r#""title":"Demo""#));
         Ok(())
--- a/crates/coven-cli/src/daemon.rs
+++ b/crates/coven-cli/src/daemon.rs
@@ -1469,11 +1469,7 @@ pub(crate) fn read_windows_pipe_http_response(
     std::thread::Builder::new()
         .name("coven-pipe-response".into())
         .spawn(move || {
-            let result = read_http_response_with_deadline(
-                &mut stream,
-                Duration::from_secs(24 * 60 * 60),
-                max_body_bytes,
-            );
+            let result = read_http_response_with_deadline(&mut stream, timeout, max_body_bytes);
             let _ = tx.send(result);
         })
         .context("failed to spawn Windows pipe response reader")?;
@@ -2440,7 +2436,10 @@ pub fn serve_forever(coven_home: &Path, started_at: String, tcp_addr: Option<&st
         local_socket::{prelude::*, GenericNamespaced, ListenerOptions},
         os::windows::local_socket::ListenerOptionsExt,
     };
-    use std::sync::Arc;
+    use std::sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    };
 
     let _ = tcp_addr; // TCP not wired on Windows in this prototype
 
@@ -2469,6 +2468,8 @@ pub fn serve_forever(coven_home: &Path, started_at: String, tcp_addr: Option<&st
         coven_home.to_path_buf(),
     ));
 
+    const MAX_INFLIGHT: usize = 64;
+    let inflight = Arc::new(AtomicUsize::new(0));
     for conn in listener.incoming() {
         let stream = match conn {
             Ok(s) => s,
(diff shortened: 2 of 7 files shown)


CompleteDotTech:
@bunsdev Rebase update: conflicts in daemon.rs, main.rs, and pty_runner.rs are resolved against current main (c409660). I preserved #315's upstream stream-mode context while retaining the Windows piped Codex lifecycle and Cave response routing. Post-rebase Windows smoke: exit 0, exactly one REBASE_CAVE_OK response, successful result event, and zero stderr bytes. GitHub now reports the PR mergeable; CI is running.

CompleteDotTech:
CI fix pushed in 647393c: restored #315's captured-output path on non-Windows, retained the Windows piped Codex/Cave path, gated platform-only helpers for Ubuntu Clippy, and updated the stale Windows extended-path assertion. Exact failing Windows test now passes; Windows Cave smoke returns CI_FIX_OK once with exit 0 and empty stderr. Replacement Ubuntu/Windows Rust jobs are currently running.

BunsDev:
@copilot Fix the code for all comments in [this review thread](https://github.com/OpenCoven/coven/pull/336#pullrequestreview-4677602221).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review thread.

CompleteDotTech:
Addressed all four review threads in 099a1a9: Windows stream-json now wraps the captured Codex transcript in JSON output events, pipe response readers use the caller timeout internally, Windows handlers are capped at 64 with inline backpressure, and the framing timing test has wider CI-safe margins. Focused Windows framing coverage passes; fmt/diff checks and tracked-tree secret scan are clean. @copilot please review the updated changes again.

BunsDev (approved):
Reviewed the latest Windows Cave/Codex reliability changes. I found no blocking issues in the current diff.

Verified locally in an isolated worktree:
- `cargo fmt --check`
- `cargo test -p coven-cli --bin coven daemon::tests::serve_lock_is_exclusive_and_reusable --locked -- --exact --nocapture`

Notes: the first full `cargo test --workspace --locked` run hit a transient `serve_lock_is_exclusive_and_reusable` lock reacquire failure; the exact test then passed, and the subsequent full locked workspace test passed. GitHub checks are green and the previously actionable Copilot threads are resolved.

PR #340, opened 2026-07-12
fix: tolerate legacy familiar records without descriptions

CompleteDotTech:
## Summary

Fix legacy `familiars.toml` records that omit `description`.

## Root Cause

Older Cave writer paths could omit `description`. Coven treated the field as required, then an API error escaped the Windows named-pipe transport and Cave displayed a roster retry state.

## Changes

- Default missing descriptions to an empty string while preserving legacy roster fields.
- Convert daemon request failures into structured JSON `500` responses.
- Cover both legacy `200` roster and malformed `500` paths over real Windows named pipes.

## Validation

- `cargo fmt --check`
- Targeted Windows named-pipe legacy roster test
- Targeted Windows malformed-registry tests
- Source-level Cave UI loaded Ash and Briar from a legacy description-less registry; human verified.

## Limitation

`cargo clippy --workspace --all-targets -- -D warnings` remains blocked by the existing `daemon_lifecycle_lock_path` dead-code warning at `daemon.rs:852`, reproduced on upstream base `a478894`.

### DIFF
--- a/crates/coven-cli/src/cockpit_sources.rs
+++ b/crates/coven-cli/src/cockpit_sources.rs
@@ -56,6 +56,9 @@ struct FamiliarEntry {
     /// treat anything else as an emoji literal.
     icon: Option<String>,
     role: String,
+    // Older CovenCave versions omitted this field for blank descriptions.
+    // Keep those registries readable while preserving a stable string in the API.
+    #[serde(default)]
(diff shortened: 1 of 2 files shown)


CompleteDotTech:
Superseded by a narrower Cave-side description requirement.

PR #341, opened 2026-07-12
fix(windows): bridge Codex stream JSON over pipes

CompleteDotTech:
## Summary

Fixes CovenCave's Windows Codex no-reply failure.

The released CLI launched npm's `codex.cmd` through the ConPTY/OpenConsole path. The wrapper could stall before Codex started, leaving CovenCave with a running session and no terminal response.

This change:

- routes `coven run codex --stream-json` through one-shot `codex exec --json` over ordinary pipes;
- passes Windows prompts through stdin (`codex exec -`) so multiline familiar prompts never enter a batch-command argument;
- normalizes Codex JSONL agent messages into Coven `assistant` frames and emits a terminal `result`;
- stores the native Codex thread id separately from Coven's stable ledger id, so callers can keep using the ledger id with `--continue`;
- bounds silent/blocked launches, protocol errors, and Unix cancellation; failed protocol turns now persist a nonzero exit code;
- cleans up detached Unix process-group descendants and uses the existing Windows Job Object / `taskkill /T /F` fallback supervision;
- updates stream, Codex, CLI, persistence, troubleshooting, and release-note documentation.

## Before / after evidence

Before, the installed Windows CovenCave process tree used the released npm CLI and followed `coven.exe -> OpenConsole.exe`, without starting Codex.

After, the patched Windows binary produced a real JSONL `assistant` response and terminal success result, then resumed the same native thread through the stable Coven ledger id. A fresh native Tauri dev instance was launched with `COVEN_BIN` set to the patched target (not via `pnpm dev:app`), and human verification confirmed both a multiline Briar reply and retained follow-up context.

## Validation

- `git diff --check`
- `cargo test -p coven-cli --quiet` on Linux
- focused Linux JSON bridge, timeout, closed-pipe descendant, zero-exit protocol-error, and SIGTERM cancellation tests
- `cargo clippy -p coven-cli --all-targets -- -D warnings`
- Windows `cargo check -p coven-cli`
- Windows `stream_json_integration`: npm-style `.cmd`, multiline stdin, assistant/result, native-thread resume, and timeout/failed-ledger regressions
- real Windows Codex smoke and stable-ledger resume against the patched `coven.exe`
- native Windows Tauri human verification

## Caveat

When Windows Job Object assignment is unavailable, Coven's own timeout/error path uses `taskkill /T /F`. An unrelated external forced termination cannot run that fallback; guaranteed forced-stop tree cleanup in that case would require a CovenCave-side process-tree kill.

### DIFF
--- a/crates/coven-cli/Cargo.toml
+++ b/crates/coven-cli/Cargo.toml
@@ -47,7 +47,7 @@ libc = "0.2"
 [target.'cfg(windows)'.dependencies]
 interprocess = "2"
 widestring = "1"
-windows-sys = { version = "0.61", features = ["Win32_Foundation", "Win32_System_Threading", "Win32_System_JobObjects"] }
+windows-sys = { version = "0.61", features = ["Win32_Foundation", "Win32_Security", "Win32_System_Threading", "Win32_System_JobObjects"] }
 
 [dev-dependencies]
 tempfile = "3"
--- a/crates/coven-cli/src/harness.rs
+++ b/crates/coven-cli/src/harness.rs
@@ -3,7 +3,7 @@ use std::env;
 use std::fs;
 use std::path::{Path, PathBuf};
 
-use anyhow::{anyhow, Result};
+use anyhow::{anyhow, Context, Result};
 use coven_runtime_spec::{Capabilities, SandboxMapping, StreamArgs};
 use serde::{Deserialize, Serialize};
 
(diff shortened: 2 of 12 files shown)


BunsDev (approved):
Reviewed the bridge end to end: structural `--json` insertion and `--`-guarded prompt are injection-safe, Windows stdin transport keeps multiline prompts off the batch command line, `harness_session_id` is additive/serde-compatible, the ledger↔native-thread mapping round-trips, and the signal handling is async-signal-safe. Local gates (fmt, clippy `-D warnings`, full workspace tests, secret scan) pass on macOS.

Pushed one commit with two fixes found in review (maintainer-edit):

1. **Busy-spin in the supervisor loop** — once all pipe reader threads exit, `recv_timeout` on the disconnected channel returns immediately, and the `Disconnected` arm was a no-op. A child that closes its stdio while still running (e.g. daemonizing wrapper) spun `try_wait` at 100% CPU until the 5-minute activity timeout. Measured 1556ms CPU over a 2s turn before the fix, 4ms after. Now sleeps for the poll interval when disconnected.

2. **Deterministic macOS test failure** — `codex_json_runner_times_out_and_reaps_a_silent_child` gave the fixture 25ms to write `child.pid` before the group kill; cold `sh` startup on macOS (~180ms measured) loses that race, failing with ENOENT (5/5 locally; CI only runs Rust checks on ubuntu/windows so it never saw it). Bumped to the 1s budget sibling tests use.

Nice work on this — the evidence-driven PR description and the bounded-failure design made it straightforward to verify. Merging once CI is green on the new head.

PR #345, opened 2026-07-12
test(harness): guard adapter-env reads with env_lock to fix tempdir race

### LINKED ISSUE #344: Flaky test: dash_prefixed_prompts_stay_positional_behind_double_dash races on adapter-manifest tempdir
## Failure

`harness::tests::dash_prefixed_prompts_stay_positional_behind_double_dash` failed on `Rust checks (ubuntu-latest)` in PR #343's CI run ([job log](https://github.com/OpenCoven/coven/actions/runs/29213000034/job/86703775142)):

```
Error: failed to read harness adapter manifest /tmp/.tmpEfOM9t/adapters/codex-compatible.json: No such file or directory (os error 2)
```

Same commit: macOS matrix leg passed, local `cargo test --workspace --locked` passed (985/985 there), and main was green at the PR's exact base (`5265f7f`). PR #343 touches only `store.rs`/`daemon.rs`/`api.rs` — nothing in `harness.rs` or the adapter-manifest path.

## Why it's an isolation race, not a code regression

The test resolves a harness adapter manifest from a `/tmp/.tmpXXXX/adapters/` tempdir that no longer exists at read time. `cargo test` runs tests in the same binary concurrently — a sibling test (or shared fixture teardown) that owns/removes that tempdir can race any test reading from it. Adding tests to the binary (as #343 did) reshuffles the parallel schedule, which is exactly the kind of perturbation that surfaces a latent race like this — the flake will keep biting unrelated PRs whenever the schedule lands wrong.

## Suggested fix directions

- Give the test its own tempdir-scoped adapter manifest (owned, not shared) so no sibling teardown can remove it mid-read, or
- If a shared fixture is intentional, hold it in a `OnceLock`/lazy static whose lifetime outlives the test run instead of a scope-dropped `TempDir`, or
- As a stopgap, mark the affected tests `#[serial]` around the shared adapter-dir resource.

Found while landing #343 (the created-row reaper); filed separately since the race predates that change.

BunsDev:
Closes #344.

## Problem

`harness::tests` mutate process-global adapter env vars (`COVEN_HARNESS_ADAPTER_DIRS` / `_MANIFEST`) under a shared `env_lock()` mutex — but **22 tests that only *read* spec resolution** (`command_parts_for_harness`, the built-in spec builders) ran without taking the lock. `cargo test` runs a module's tests concurrently, so a reader could observe a sibling test's half-torn-down tempdir manifest and blow up:

```
Error: failed to read harness adapter manifest
/tmp/.tmpXXXX/adapters/codex-compatible.json: No such file or directory (os error 2)
```

This is exactly what reddened CI on #343 — a daemon change that touched no harness code but reshuffled the in-binary test schedule enough to surface the latent race. (Confirmed benign there: main went green on the merge commit `8df3e610`.)

## Fix

Take `env_lock().lock()` at the top of every spec-resolving test, so reads serialize against the mutators that install/remove the tempdir manifests. Test-only change; **no production code touched** (diff is `crates/coven-cli/src/harness.rs` tests module only).

## Validation

- `cargo fmt --check` ✓ · `cargo clippy --workspace --all-targets -- -D warnings` ✓ · `cargo test --workspace --locked` ✓ · secret scan (pre-commit gitleaks) ✓
- **Stress:** ran `cargo test -p coven-cli harness::` 5× in a row — 55/55 each time, no flake.
- Verified no test ends up double-locking the non-reentrant mutex (per-function guard count audited: 31 tests, exactly one guard each).

### DIFF
--- a/crates/coven-cli/src/harness.rs
+++ b/crates/coven-cli/src/harness.rs
@@ -1570,6 +1570,9 @@ mod tests {
 
     #[test]
     fn command_parts_for_known_harnesses_append_interactive_prompt() -> anyhow::Result<()> {
+        // Spec resolution reads the adapter env vars; hold the shared env
+        // lock so a concurrent test's tempdir manifest never vanishes mid-read.
+        let _guard = env_lock().lock().unwrap();
         assert_eq!(
             command_parts_for_harness("codex", "fix tests", HarnessLaunchMode::Interactive)?,
             (
@@ -1589,6 +1592,9 @@ mod tests {
 
     #[test]
     fn command_parts_for_known_harnesses_use_noninteractive_entrypoints() -> anyhow::Result<()> {
+        // Spec resolution reads the adapter env vars; hold the shared env
+        // lock so a concurrent test's tempdir manifest never vanishes mid-read.
+        let _guard = env_lock().lock().unwrap();
         assert_eq!(
             command_parts_for_harness("codex", "fix tests", HarnessLaunchMode::NonInteractive)?,
             (
@@ -1619,6 +1625,9 @@ mod tests {
 
     #[test]
     fn dash_prefixed_prompts_stay_positional_behind_double_dash() -> anyhow::Result<()> {
+        // Spec resolution reads the adapter env vars; hold the shared env
+        // lock so a concurrent test's tempdir manifest never vanishes mid-read.
+        let _guard = env_lock().lock().unwrap();
         // A prompt starting with `-` must never parse as harness flags.
         for mode in [
             HarnessLaunchMode::Interactive,
@@ -1677,6 +1686,9 @@ mod tests {
 
     #[test]
     fn command_parts_reject_unknown_harnesses() {
+        // Spec resolution reads the adapter env vars; hold the shared env
+        // lock so a concurrent test's tempdir manifest never vanishes mid-read.
+        let _guard = env_lock().lock().unwrap();
         assert!(
             command_parts_for_harness("shell", "hello", HarnessLaunchMode::Interactive)
                 .unwrap_err()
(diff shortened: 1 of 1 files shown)


BunsDev:
## Review — verified, ready to merge ✅

Went beyond the diff and validated the fix against the full test module on this branch:

**Correctness of the guard placement**
- All **31** `env_lock()` sites now cover every test that resolves specs through `command_parts_for_harness*` / `configured_harness_specs` — i.e. every reader of `COVEN_HARNESS_ADAPTER_MANIFEST` / `COVEN_HARNESS_ADAPTER_DIRS`.
- Every env **writer** test (6 `set_var`/`remove_var` sites outside the restore helpers) already held the lock before this PR, so with readers now guarded the #344 window (writer restores the tempdir manifest mid-read) is fully closed.
- The two spec-reading tests deliberately left unguarded (`built_in_harnesses_returns_codex_and_claude`, `built_in_harnesses_include_first_run_recovery_commands`) call `built_in_harness_specs()`, a pure constructor with no env read — correctly exempt.

**No new hazards**
- No deadlock risk: `env_lock` is `cfg(test)`-only and production code never locks, so no guarded test can re-enter the mutex through the functions under test.
- `.lock().unwrap()` matches the file's existing convention. *Optional hardening (non-blocking, fine as a follow-up):* a panicking guarded test will poison the mutex and cascade `PoisonError` failures across the remaining ~30 tests, masking the real failure — `env_lock().lock().unwrap_or_else(std::sync::PoisonError::into_inner)` would keep the innocent tests green. Only worth doing if you touch this file again.

CI on head `5dad77c` is fully green. Fixes the #344 flake at its root rather than retrying/serializing the whole suite. LGTM.

PR #346, opened 2026-07-12
feat(cli): unify coven-code as a managed engine (Phase 0+1)

BunsDev:
## Summary

Phase 1 of unifying the OpenCoven CLIs: make **`coven` the single user-facing CLI**, driving `coven-code` (the "engine") as a **separate managed process — never linked**. This is the "one surface, two processes" model.

**Why a process boundary, not one binary:** `coven` is MIT; the engine is GPL-3.0 (inherited from upstream Claurst, not unilaterally relicensable). Statically linking the engine's crates would force the combined binary to GPL. Invoking the engine at arm's length as a subprocess keeps each work under its own license — the ripgrep-in-VS-Code / containerd-under-Docker pattern. This PR adds **no new dependencies** and **no dependency on any `claurst-*` crate**.

Full design + all 6 phases: `docs/superpowers/plans/2026-07-12-coven-cli-unification.md` (in the coven-code repo). The compatibility contract is `docs/ENGINE-CONTRACT.md` here.

## What's in this PR (Phase 0 + Phase 1)

- **`docs/ENGINE-CONTRACT.md`** (contract_version 1) — the exact CLI/env/stream surfaces coven relies on, each verified against the real engine binary.
- **`crates/coven-cli/src/engine.rs`** — managed-engine resolver: `COVEN_ENGINE_BIN` → managed `~/.coven/engine/current` → PATH → legacy `~/.coven-code/bin`; version gate (`MIN_ENGINE_VERSION` 0.6.1) with a `--version` handshake at delegation.
- **`crates/coven-cli/src/engine_install.rs`** — `coven engine install`: downloads the platform release archive (`.tar.gz`/`.zip`) via `curl`, verifies the archive SHA-256 **before** extraction (fails closed on mismatch), extracts via `tar`, atomically installs into `~/.coven/engine/<ver>/`. RAII scratch-cleanup on every exit.
- **`coven engine status | install | which`**, **first-run auto-install prompt** (TTY-gated, `COVEN_NO_AUTO_INSTALL` opt-out), **passthroughs** `coven auth | models | acp` and the raw `coven code` escape hatch, and a **doctor `Engine:` section** (source, version-vs-minimum, pin placeholder, auth via a 5s-bounded `auth status --json` probe — auth is non-blocking).

## Behavior

- `npm i -g @opencoven/cli && coven` → offers to install the engine, then opens the TUI. Zero separate coven-code install step.
- Delegation forwards `COVEN_PARENT=coven` (+ `COVEN_HOME` when set) and refuses engines older than the minimum with an actionable message.
- The managed engine wins resolution over a PATH copy, so `coven engine install` is authoritative.

## Verification

- `cargo test -p coven-cli` — **999 + 4 + 4 + 20 + 1 passed, 0 failed** (rebased onto current `main`).
- `cargo clippy -p coven-cli --all-targets -- -D warnings` — clean.
- `cargo fmt --check` — clean.
- **Live smoke:** `coven engine install --version 0.6.1` downloads + extracts + runs (`coven-code 0.6.1`); `coven code --version`, `coven auth status --json`, `coven models` reach the engine; `coven doctor` shows the Engine section and exits 0 with auth logged-out.

## Review process

Each task was implemented TDD-style and passed a two-stage review (spec compliance + code quality) with fixes looped back in; a final whole-branch review confirmed the license boundary, cross-task consistency (single-sourced `MIN_ENGINE_VERSION`/`DEFAULT_ENGINE_VERSION`, install-writes-where-resolve-reads), and commit hygiene. All 8 commits are signed.

## Not in this PR (later phases)

Pinned `engine.lock` + cross-repo contract-test CI (Phase 2), engine-as-first-class-harness with daemon-ledgered sessions (Phase 3), state/config/auth unification under `~/.coven` (Phase 4), brand/UX sweep (Phase 5). Pin placeholders are honest `// Task 2.1` markers, not half-built.

## Companion

The engine-side contract pointer (COVEN.md + AGENTS.md) is a small separate PR against `OpenCoven/coven-code`.

### DIFF
--- a/crates/coven-cli/src/engine.rs
+++ b/crates/coven-cli/src/engine.rs
@@ -1,0 +1,297 @@
+//! Managed engine (coven-code) resolution and version gating.
+//!
+//! LICENSE BOUNDARY: the engine is GPL-3.0; coven is MIT. The engine is
+//! always a separate process launched by path — never a Cargo dependency.
+//! Do not add any claurst-* crate to this workspace.
+
+use anyhow::{anyhow, bail, Context, Result};
+use std::ffi::OsStr;
+use std::path::{Path, PathBuf};
+use std::process::Command;
+
+#[cfg(windows)]
+pub const ENGINE_BIN_NAME: &str = "coven-code.exe";
+#[cfg(not(windows))]
+pub const ENGINE_BIN_NAME: &str = "coven-code";
+
+/// Oldest engine this coven build can drive (contract v1 surfaces).
+// used by Task 1.2+
+#[allow(dead_code)]
+pub const MIN_ENGINE_VERSION: (u64, u64, u64) = (0, 6, 1);
+
+#[derive(Debug)]
+pub enum EngineSource {
+    EnvOverride, // COVEN_ENGINE_BIN
+    Managed,     // ~/.coven/engine/<current>/
+    PathLookup,  // coven-code on PATH
+    LegacyHome,  // ~/.coven-code/bin/ (pre-unification installs)
+}
+
+// used by Task 1.2+
+#[allow(dead_code)]
+#[derive(Debug)]
+pub struct ResolvedEngine {
+    pub path: PathBuf,
+    pub source: EngineSource,
+}
+
+pub fn resolve() -> Option<ResolvedEngine> {
+    let env_override = std::env::var_os("COVEN_ENGINE_BIN");
+    let path_var = std::env::var_os("PATH");
+    let home = dirs_next::home_dir();
+    resolve_from(
+        env_override.as_deref(),
+        path_var.as_deref(),
+        home.as_deref(),
+    )
+}
+
+pub fn resolve_from(
+    env_override: Option<&OsStr>,
+    path_var: Option<&OsStr>,
+    home: Option<&Path>,
+) -> Option<ResolvedEngine> {
+    // 1. COVEN_ENGINE_BIN explicit override
+    if let Some(override_path) = env_override {
+        let p = PathBuf::from(override_path);
+        if is_executable(&p) {
+            return Some(ResolvedEngine {
+                path: p,
+                source: EngineSource::EnvOverride,
+            });
+        }
+        // Set but not executable: fall through to next source
+    }
+
+    // 2. Managed ~/.coven/engine/<current>/ENGINE_BIN_NAME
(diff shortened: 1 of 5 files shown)


BunsDev:
Reviewed against the engine source (coven-code main) and the live CI runs.

**Verified:**

- License boundary holds: no `claurst-*` dependency, engine driven only as a subprocess (unix `exec`, exit-code propagation elsewhere).
- Resolver precedence (env override → managed → PATH → legacy) is correct and the edge cases are tested (non-executable override falls through, whitespace-only `current` file skipped).
- `engine_install` is fail-closed: archive SHA-256 verified before extraction, RAII scratch cleanup on every exit path, atomic rename for both the binary and the `current` pointer.
- Doctor's `auth status --json` probe kills and reaps the child on the 5s deadline.
- The 108 removed lines in main.rs are the old PATH+legacy lookup helper, fully superseded by `engine::resolve()`.
- ENGINE-CONTRACT.md spot-checked claim-by-claim against the engine: `--cwd`, `--permission-mode` values, `auth status --json` `loggedIn` + exit 0/1 (`main.rs:5140`), both stdin frame shapes incl. assistant prefill (`stream_mode.rs`), exit-on-EOF. All accurate.

**Blocking: Secret guard.** All 10 findings (3 current + 7 history) are false positives from two patterns in the new code:

- `crates/coven-cli/src/engine.rs` `let token = text.split_whitespace().last()?;` / `let token = token.strip_prefix(...)` — the identifier `token` plus `=` trips `generic_assignment`.
- `engine_install.rs` test: the `releases/download/...` URL trips `high_entropy` — the URL allowlist only recognizes `blob|tree` under `OpenCoven/coven/`.

Because the guard also scans branch history, fixing the tip lines cannot clear the 7 history findings. The scanner needs the fix. Note `feat/engine-lock`, `feat/engine-harness`, and `feat/engine-ledger` currently fail Secret guard the same way, so a scanner fix on main unblocks the whole engine train. PR incoming; once it lands, update this branch (or re-run CI) to go green.

**Known-transitional, fine as-is:** auto-install and `coven engine install` pass `expected_sha256: None` (loud unverified warning) until Task 2.1's engine.lock pin. Nit: the install `curl` could add `--proto '=https' --tlsv1.2`.

PR #349, opened 2026-07-13
feat(daemon): register externally-owned sessions in the ledger (Phase 3, Track B)

BunsDev:
## Summary

Phase 3, **Track B**: the coven daemon can now **register externally-owned sessions**, so an interactive engine (coven-code) TUI session shows up in `coven sessions`. Stacks on #348 (Track A) — targets `feat/engine-harness`, so the diff is only the Track B delta. The engine-side notifier is the companion PR on OpenCoven/coven-code.

This closes Phase 3's ledger goal: the daemon's existing session API *launches and owns a PTY*, but a TUI session is already running elsewhere — it must be *registered*, not launched.

## What's in this PR (daemon side)

- **`POST /api/v1/sessions/external`** — register an already-running, externally-owned session (`{id, projectRoot, harness, title?, transcriptPath?}`, camelCase). Inserts a `running` + `external` `SessionRecord` **without spawning a harness/PTY**. Idempotent (201 first, 200 re-register); 409 if the id collides with a daemon-managed session.
- **`POST /api/v1/sessions/:id/complete`** — mark it finished (`{exitCode?}`; nonzero → failed). 404 if unknown; **422 if the session isn't external** (prevents completing a live daemon-PTY session and diverging the ledger from reality).
- **Schema:** `external` + `transcript_path` columns, added to fresh DBs and via idempotent `ensure_column` migration. **Orphan recovery now skips external sessions** (`AND external = 0`) so a daemon restart doesn't wrongly reap a live engine session.
- **`kill` on an external session → 422** (`external_session_not_killable`) — the daemon doesn't own its process.
- **`attach`** on an external session shows the recorded ledger with a clear "not the live terminal" notice, and doesn't try to forward stdin to a PTY that doesn't exist.
- **Contract docs:** the two endpoints documented in `API-CONTRACT.md`, plus a "Daemon endpoints the engine calls" section in `ENGINE-CONTRACT.md`.

## Verification

- `cargo test -p coven-cli` — **1025 + 4 + 4 + 20 + 1 passed, 0 failed** (register/complete happy paths, all guards 422/422/409/404, idempotency, orphan-skip regression, schema round-trip).
- `cargo clippy --all-targets -- -D warnings` clean; `cargo fmt --check` clean.

## Review

[...]

PR #352, opened 2026-07-13
feat(doctor+search): unified Credentials panel + cross-session search (Phase 4, Track D)

BunsDev:
## Summary

Phase 4, **Track D**: one auth story and search that spans engine sessions. Two coven-side additions, stacked on #349 (Phase 3 Track B) — targets `feat/engine-ledger`, so the diff is only the Track D delta. Companion Track C (engine home/config unification under `~/.coven`) is a separate PR on OpenCoven/coven-code.

## What's in this PR

- **Unified "Credentials" doctor panel** (`bc031bb`): `coven doctor` now answers "am I logged in, to what, via what" in one place — the engine's real auth (`auth status --json` → logged in/out) plus each available harness (codex/claude) with an authenticate hint. It doesn't fabricate a logged-in status for harnesses that expose no machine-readable auth probe. The engine-auth subprocess is called once; exit-code semantics are unchanged (auth is a non-blocking warning). Pure `credentials_lines` helper with unit tests.
- **Cross-session search** (`f347467`): `coven sessions search` now finds content from interactive engine (TUI) sessions. Those register as external sessions (Phase 3) with a `transcript_path` but no daemon events; this lazily ingests their JSONL transcript into the events table on search — through the **existing privacy-aware insert path** (so it's redacted and retention-bounded like any event, and auto-indexed into FTS5). Bounded (line + byte caps with truncation warnings), idempotent (a `transcript_indexed_at` marker), and best-effort (a failing ingest never fails the search).

## Verification

- `cargo test -p coven-cli` — green; `cargo clippy --all-targets -- -D warnings` clean; `cargo fmt --check` clean.
- Both tasks passed spec + quality review: the Credentials panel confirmed no double subprocess / no exit-code change; the search ingestion confirmed redaction-not-bypassed, bounded, idempotent, best-effort, and no positional-mapper desync.

### DIFF
--- a/crates/coven-cli/src/main.rs
+++ b/crates/coven-cli/src/main.rs
@@ -751,6 +751,32 @@ fn edit_distance(a: &str, b: &str) -> usize {
 fn run_sessions_search(query: &str, json: bool) -> Result<()> {
     let store_path = coven_store_path()?;
     let conn = store::open_store(&store_path)?;
+
+    // Lazily ingest external-session transcripts (e.g. coven-code TUI sessions)
+    // so they become searchable. This is a one-time cost per session: once
+    // `transcript_indexed_at` is set the ingest function is a no-op.
+    let coven_home = coven_home_dir()?;
+    let now = chrono::Utc::now().to_rfc3339_opts(chrono::SecondsFormat::Nanos, true);
+    match store::list_uningest_external_sessions(&conn) {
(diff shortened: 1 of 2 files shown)


BunsDev:
Superseded by the clean re-landed stack (#353 → #354 → #355 → #356), all merged to `main`. Closing.

PR #334, opened 2026-07-09
feat(harness): integration PR 3 stream continuity manifests

BunsDev:
## Context

- Summary: Integration PR 3 from PR #332's handoff. This removes the foreground `coven run --stream-json` Claude id gate, moves one-shot continuity into adapter declarations, and exposes declared harness capabilities in JSON adapter summaries.
- Files changed: `crates/coven-cli/src/harness.rs`, `crates/coven-cli/src/main.rs`, `crates/coven-cli/src/pty_runner.rs`, `docs/HARNESS-ADAPTERS.md`, `docs/STREAM-JSON.md`
- Closes #

## Implementation

- Approach:
  - `HarnessSummary` now includes the declared `capabilities` block, so `coven adapter list --json` consumers can read `stream`, `preassigned_session_id`, `think`, and `speed` directly.
  - Foreground stream-json pass-through now gates on `selected_harness.capabilities.stream` and launches with argv built from declared `stream_args`; `pty_runner::stream_harness` only spawns the already-built command and forwards JSONL.
  - Added manifest-driven `continuity_args` for one-shot non-interactive init/resume behavior. Bundled Claude and Codex express their existing behavior through these declarations, and external manifests can do the same.
  - Added validation for invalid/dead continuity declarations and updated adapter/stream-json docs.
- User-visible behavior: stream-capable manifest adapters can use native foreground JSONL pass-through and one-shot continuity without Coven core id checks. Adapter JSON summaries include declared capabilities.
- Compatibility notes: legacy manifests remain valid because `continuity_args` defaults to absent. Codex and Claude argv behavior is test-pinned; non-stream harnesses continue using captured PTY `output` frames.

## Verification

- [x] `cargo fmt --all`
- [x] `cargo fmt --all -- --check`
- [x] `cargo check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test -p coven-cli --quiet` — 983 passed, 1 ignored in unit tests; integration test binaries passed.
- [x] `cargo test --workspace --locked --quiet` — workspace tests passed.
- [x] `python3 scripts/check-secrets.py` — `Secret guard passed: no current-tree or history findings.`
- [x] Additional manual checks: `git diff --check`; targeted red/green tests for continuity manifest handling, stream harness passthrough, and stream user-event synthesis.

## Risk and Rollback

[...]

PR #363, opened 2026-07-14
feat(run): repeatable --add-dir harness trust passthrough

BunsDev:
## Summary

Implements the coven side of Cave's granted-roots forwarding (bead `cave-t1ta`, companion to OpenCoven/coven-cave#2969): `coven run` now accepts a repeatable `--add-dir <DIR>` and forwards each directory to the harness's native trust flag, so granted project roots become real filesystem grants instead of advisory prompt text.

## What changed

- **CLI**: `coven run --add-dir <DIR>` (repeatable). Advertised in `run --help`, which is exactly what Cave's capability probe gates forwarding on (`covenRunSupportsAddDirFlag`).
- **Spec**: new `add_dir_flag` on `HarnessCommandSpec`; built-in codex, claude, and coven-code engine all declare `--add-dir` (verified against each installed binary: `codex exec --add-dir`, `claude --add-dir`, `coven-code --add-dir`).
- **Arg assembly**: add-dir pairs ride ahead of the prompt positional alongside model/sandbox args in every launch mode (one-shot, continuity, stream), plus the dedicated claude stream-json path.
- **External adapters**: manifests declare `add_dir_flag` / `addDirFlag`; adapters that declare none make `--add-dir` a warned no-op (mirrors `--model` / `--permission` degradation).
- **Docs**: README run table + `docs/HARNESS-ADAPTERS.md` field reference.

## Testing

- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings` clean
- `cargo test --workspace --locked`: 1048/1048 coven-cli tests pass (note: run with `COVEN_HARNESS_ADAPTER_DIRS` unset; a stale `coven-code.json` external adapter in that dir conflicts with the new built-in engine id and fails ~40 pre-existing tests regardless of this change)
- `python3 scripts/check-secrets.py` clean
- New tests: codex/claude forwarding + ordering, blank no-op, spec-without-mechanism no-op, built-ins-declare-flag, manifest `addDirFlag` parse, claude stream-json forwarding, repeatable clap parse, and a `run --help` render test asserting the probe contract.
- Manual: `./target/debug/coven run --help` matches Cave's probe regex `/(^|\\s)--add-dir(?![\\w-])/m` → `true`.

## Readiness

- Scope: CLI flag + spec plumbing only; no daemon API surface change.
- Risk: low — absent flag preserves existing argv byte-for-byte (empty-args early return).
- Follow-up: Cave side already merged and probe-gated, so this activates forwarding on release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,6 +260,7 @@ Choose a repo, choose a harness, get a verified patch.
 | `coven run <harness> <prompt> --model <id>`    | Forward a model override to the harness                          |
 | `coven run <harness> <prompt> --think`         | Request deeper reasoning when the harness supports it            |
 | `coven run <harness> <prompt> --speed <level>` | Set a latency/reasoning hint: `fast`, `balanced`, or `thorough`  |
+| `coven run <harness> <prompt> --add-dir <dir>` | Trust an additional directory (repeatable); maps to the harness's native `--add-dir` flag |
 | `coven run <harness> --continue`               | Resume the latest active session for this project                |
 | `coven run <harness> --continue <id>`          | Resume a specific session by id                                  |
 | `coven run <harness> <prompt> --detach`        | Create the session record without launching the harness          |
--- a/crates/coven-cli/src/harness.rs
+++ b/crates/coven-cli/src/harness.rs
@@ -121,6 +121,12 @@ pub struct HarnessLaunchOptions<'a> {
     /// sandbox flag when the spec declares one. `Option<Permission>` stays
     /// `Copy` because `Permission` is `Copy`.
     pub permission: Option<Permission>,
+    /// Additional directories the harness should trust beyond its cwd
+    /// (`coven run --add-dir <DIR>`, repeatable). Each entry forwards as
+    /// `[add_dir_flag, <dir>]` when the spec declares an add-dir mechanism;
+    /// harnesses that declare none make the flag a warned no-op. Blank
+    /// entries are skipped. A shared slice keeps the struct `Copy`.
+    pub add_dirs: &'a [String],
 }
 
 impl<'a> HarnessLaunchOptions<'a> {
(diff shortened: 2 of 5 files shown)


BunsDev:
## Review — verified, ready to merge ✅

Reviewed with the trust boundary as the focus, since `--add-dir` widens what the harness may touch:

**Trust model holds**
- The grant is **explicit and per-run, CLI-only**: `add_dirs` flows `Run` → `run_session` → `HarnessLaunchOptions` and is *not* exposed through `POST /api/v1/sessions` — clients can't widen access silently; Cave has to shell through `coven run` (hence the help-probe test). That matches how `--permission` is handled and keeps the daemon boundary intact.
- Warned no-op for adapters that don't declare `add_dir_flag` mirrors the `model_flag`/`sandbox` degradation pattern exactly, including the `if let Some(s) = spec.as_ref()` guard (spec is always `Some` for a validated harness, same as the model warning path).

**Flag mechanics verified against the real CLIs**
- Confirmed locally: `codex --help` and `codex exec --help` both accept `--add-dir <DIR>` ("Additional directories that should be writable alongside the primary workspace"), so the root-position ordering pinned by `codex_forwards_add_dirs_before_prompt` is valid.
- Claude stream path (`pty_runner`) forwards the same repeated pairs inline and the fake-claude argv test pins the exact final argv including blank-entry skipping — good end-to-end coverage of both launch paths.
- Blank filtering happens in both `run_session` and `add_dir_args` — redundant but harmless (the runner path relies on the second).

**Nits (non-blocking)**
- `run_session` trims/filters into `requested_add_dirs`, then `add_dir_args` trims/filters again; if you ever touch this again, one layer can own it (the arg builder, since the stream path also uses per-entry trimming).
- Heads-up for sequencing with #334: both PRs extend the adapter-manifest schema and the `HARNESS-ADAPTERS.md` capability table; whichever lands second has a small doc/struct-literal rebase (all `HarnessCommandSpec` literals gain a field).

CI green on head `1105fb9`. LGTM.

PR #367, opened 2026-07-14
feat(cli): observability parity with Cave and the daemon API (status, familiars, skills, memory, research, calls, hub, session inspection)

### LINKED ISSUE #366: CLI observability parity: Cave/API surfaces (overview, familiars, skills, memory, research, calls, hub, session detail) unreachable from the CLI
## Problem

The daemon API (and CovenCave on top of it) exposes a set of user-relevant read surfaces that the CLI/TUI cannot reach today. A terminal-only user (or a script) has no way to see what Cave sees without hand-crafting `curl --unix-socket` calls — which is exactly what `docs/HUB-OPERATIONS.md` currently tells operators to do.

### Parity audit (API ↔ CLI)

| Surface | API route | CLI today |
|---|---|---|
| Ecosystem overview | `GET /api/v1/overview` | ❌ none (and the endpoint returns hardcoded zero counts) |
| Familiars roster | `GET /api/v1/familiars` | ❌ none |
| Skills inventory | `GET /api/v1/skills` | ❌ none |
| Memory files | `GET /api/v1/memory` | ❌ none |
| Research log | `GET /api/v1/research` | ❌ none |
| Coven Calls ledger | `GET /api/v1/coven-calls[/:id]` | ❌ none |
| Hub status/nodes/jobs/routing | `GET /api/v1/hub/*` | ❌ none — docs say use `curl --unix-socket` |
| Session detail | `GET /api/v1/sessions/:id` | ❌ no non-interactive inspect |
| Session events | `GET /api/v1/sessions/:id/events` (`afterSeq`, `limit`) | ❌ only interactive replay via `coven attach` |
| Session log | `GET /api/v1/sessions/:id/log` | ❌ only interactive replay |

### Related UX gaps

- `GET /api/v1/overview` returns hardcoded `0` for familiar/skill/memory/research counts, so Cave's overview shows zeros even when `~/.coven` is populated.
- README Commands Reference omits existing commands: `kill`, `vacuum`, `completions`, `sessions search`, `claim release|heartbeat|canary`.
- `docs/reference/cli.md` front-matter mentions a `view` command that does not exist.

## Proposal

Add read-only CLI commands that render through the same in-process API handler (byte-for-byte `--json` parity, no daemon required for store/file-backed reads, matching how `coven sessions` reads the store directly):

- `coven status` (alias `overview`) — daemon health + session counts + familiars/skills/memory/research counts + hub summary
- `coven familiars|skills|memory|research|calls [--json]`
- `coven hub status|nodes|jobs|routing [--json]` (read-only; write paths stay machine-only)
- `coven sessions show|events|log <id>` with `--json`, `--after-seq`, `--limit`
- Fix `overview` endpoint to compute real counts
- Docs: README commands reference, reference pages, HUB-OPERATIONS curl → CLI

Out of scope: travel/scheduler/actions/executor write paths (machine-to-machine protocol surfaces), TUI chat panes, `PUT /familiars/:id/icon`.

BunsDev:
## Context

- Summary: Every read surface CovenCave renders — overview, familiars, skills, memory, research, Coven Calls, hub control-plane, session detail/events/log — is now reachable from the CLI, for humans (tables with teaching empty states) and machines (`--json` carrying the same body as the daemon API route). Also fixes `GET /api/v1/overview` returning hardcoded zero counts, so Cave's dashboard benefits too.
- Files changed: `crates/coven-cli/src/observe.rs` (new), `crates/coven-cli/src/main.rs`, `crates/coven-cli/src/api.rs`, README, `docs/reference/cli-observe.md` (new), `docs/reference/cli.md`, `docs/reference/cli-sessions.md`, `docs/HUB-OPERATIONS.md`, `docs/GETTING-STARTED.md`, design spec under `docs/superpowers/specs/`.
- Closes #366

## Implementation

- Approach: new commands render through the **in-process API router** (`api::handle_request`), so `--json` output is the daemon API body for the same route (pretty-printed) and the human tables are a rendering of that same body — one source of truth for shapes, no daemon required for these offline reads (same pattern as `coven sessions` reading the store directly). Full parity matrix, alternatives, and devil's-advocate notes are in the committed design spec.
- User-visible behavior:
  - `coven status` (alias `overview`) — daemon, sessions, familiars, skills, research, hub at a glance; first-run empty states teach next steps; `--json` is a documented `{health, overview}` composition (precedent: `coven daemon status --json`).
  - `coven familiars | skills | memory | research [--json]`
  - `coven calls [<id>] [--json]` — Coven Calls delegation ledger
  - `coven hub status|nodes|jobs|routing [--json] [--state <s>]` — hub ops without hand-rolled `curl --unix-socket` (HUB-OPERATIONS runbook now leads with the CLI)
  - `coven sessions show|events|log <id>` — non-interactive session inspection, unique-prefix id resolution, `--after-seq`/`--limit` cursor paging matching the API envelope
  - `GET /api/v1/overview` now computes real counts (`active_familiars` = distinct roster familiars with an open session, so it can never exceed `total_familiars`); unreadable side sources degrade to 0 (dashboard semantics) while leaf routes still fail loudly
  - Top-level `--help` gains a four-line "Common first steps" footer; near-miss typo suggestions cover the new commands automatically
- Compatibility notes: additive CLI surface; no API shape changes (overview keeps its exact field set — real numbers only). New top-level words (`status`, `familiars`, `skills`, `memory`, `research`, `calls`, `hub`) shadow *unquoted* multi-word cast prompts starting with those words — same tradeoff the existing `kill`/`patch`/`run` words already made; quoted prompts (`coven "memory leak…"`, the documented cast form) are unaffected.

## Verification

- [x] `cargo fmt --check`
- [x] `cargo test --workspace --locked` (1070+ tests; re-run after rebase onto #341)
- [x] `python3 scripts/check-secrets.py` (current tree + history clean)
- [x] Additional manual checks: smoke-tested against an isolated `COVEN_HOME` — empty first-run states, populated roster/skills/memory/research/calls, hub single-host + registered-node states, `sessions show/events/log` against a `--detach` session including unique-prefix and not-found error paths; verified quoted cast prompts still parse as prompts; verified `--help`/`-h` output.

## Risk and Rollback

- Risk level: Low — read-only command surface plus one endpoint count fix. The riskiest piece is the overview computation; it degrades to zeros on unreadable sources rather than erroring, covered by two endpoint tests.
- Rollback plan: revert the branch commits; no migrations, no protocol changes, no client contract changes.

## Agent Handoff

- Current state: complete and green; each command has renderer unit tests plus end-to-end parity tests that feed live API bodies from seeded temp homes into the renderers (catches DTO key drift — this caught the `research/results.tsv` path and the bare-array log envelope during development).
- Follow-ups:
  - The `GET /api/v1/capabilities` route literal arm shadows the later `crate::capabilities::get_all` arm (harness capability aggregate is unreachable through the router; `?refresh=1` handling is dead there). Left untouched because changing that route's body is a client-visible behavior change — will file as its own issue.
  - Skill `score`/rate fields are stubbed at 0.0 upstream; `average_skill_score` honestly reports 0 until scoring lands.
  - TUI chat panes for these views are deliberately out of scope (engine surface).
- Known gaps: `coven status --json` composes two API bodies rather than proxying a single route — documented as CLI-owned in `docs/reference/cli-observe.md`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,8 @@ Choose a repo, choose a harness, get a verified patch.
 | `coven chat`     | Explicitly open the interactive Coven UI                        |
 | `coven tui`      | Same as `coven chat`                                            |
 | `coven doctor`   | Detect supported harness CLIs and print install hints           |
+| `coven status`   | Ecosystem overview: daemon, sessions, familiars, skills, hub (alias: `coven overview`) |
+| `coven completions <shell>` | Generate shell completions (bash, zsh, fish, elvish, powershell) |
 
 ### Harness adapters
 
@@ -275,13 +277,19 @@ Choose a repo, choose a harness, get a verified patch.
 | `coven sessions --plain`                       | Force plain table output for scripts or copying                  |
 | `coven sessions --json`                        | Output sessions as JSON                                          |
 | `coven sessions --json --all`                  | Output all sessions (including archived) as JSON                 |
+| `coven sessions search <query>`                | Full-text search recorded event payloads (`--json` for scripts)  |
+| `coven sessions show <session-id>`             | Show one session's record without attaching (`--json` for scripts) |
+| `coven sessions events <session-id>`           | List recorded events; `--after-seq`/`--limit` page the cursor    |
+| `coven sessions log <session-id>`              | Print a session's log lines without attaching                    |
 | `coven attach <session-id>`                    | Replay/follow session output and forward input                   |
 | `coven summon <session-id>`                    | Restore an archived session, then replay/follow it               |
 | `coven archive <session-id>`                   | Hide a non-running session while preserving its events           |
 | `coven sacrifice <session-id> --yes`           | Permanently delete a non-running session and its events          |
+| `coven kill <session-id>`                      | Kill a running session's process (its event log is kept)         |
 
-> `attach`, `summon`, `archive`, and `sacrifice` accept a unique prefix of the
-> session id (e.g. `coven attach 9099`), so you don't have to paste full UUIDs.
+> `attach`, `summon`, `archive`, `sacrifice`, `kill`, and the `sessions
+> show/events/log` subcommands accept a unique prefix of the session id
+> (e.g. `coven attach 9099`), so you don't have to paste full UUIDs.
 
 > **Session rituals are intentionally explicit.** Archive is reversible and keeps the full event ledger. Summon brings an archived session back. Sacrifice is destructive, refuses live sessions, and requires `--yes` so beginners don't delete work by accident.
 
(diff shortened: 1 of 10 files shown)


BunsDev:
Verified the reviewer's three follow-up commits locally on top of the branch: `cargo fmt --check`, `cargo clippy -p coven-cli --all-targets -- -D warnings`, and the full bin suite (1078 passed) are green. The u64 tightening on `--after-seq`/`--limit` is correct — clap now rejects negatives at parse time instead of forwarding them into the API query string — and the docs/spec edits fix real drift (status help omitted research; spec described the pre-implementation active_familiars semantics). CI is re-running on the new head.

PR #370, opened 2026-07-14
feat(tui): observe views in Cast and chat — /status, /familiars, /skills, /memory, /research, /calls, /hub

BunsDev:
## Context

- Summary: Extends #367's observability parity to the **interactive surfaces**: the Cast composer and the chat UI can now open the same read-only views (`/status`, `/familiars`, `/skills`, `/memory`, `/research`, `/calls`, `/hub`) that the CLI commands render, through the same read path, so the three surfaces can never disagree. Also fixes a consistency bug #367 introduced: the Cast shell's plain word `status` used to run *doctor*, which now contradicted `coven status`.
- Files changed: `crates/coven-cli/src/observe.rs`, `crates/coven-cli/src/tui/cast/{intent,plan,mod}.rs`, `crates/coven-cli/src/tui/shell.rs`, `crates/coven-cli/src/tui/chat/app.rs`, `crates/coven-cli/src/main.rs` (palette test pins), `CHAT-QUICKSTART.md`, `docs/reference/cli-observe.md`.
- **Stacked on #367** (base branch `feat/366-cli-observability-parity`; retarget to `main` + rebase after it lands). Refs #366.

## Implementation

- Approach: one new `CastIntent::Observe { view: ObserveView }`. `ObserveView` lives in `observe.rs` (the data layer) and is re-exported by `cast::intent`, keeping layering one-directional (tui → observe). A shared `observe::view_text(coven_home, view)` renders the human text for all surfaces; the CLI human path now goes through it too (no second render path).
- User-visible behavior:
  - Cast (`coven "<spell>"` and the composer): `/status` (alias `/overview`), `/familiars`, `/skills`, `/memory`, `/research`, `/calls`, `/hub`, plus bare words (`status`, `overview`, `familiars`, `roster`, `skills`, `memory`, `research`, `calls`, `hub`, `hub status`). Plan cards are Safe/Inform and name the scriptable `coven <view>` command.
  - Chat UI: views render as system messages, followed by a "Scriptable form: `coven <view>` (add --json for machines)" hint; read failures degrade to a hint message, never a crash.
  - Legacy palette (behind `COVEN_LEGACY_TUI=1`): one new "Coven status" entry (key `s`, `/status`).
  - `status`/`overview` in a composer now mean the ecosystem overview (CLI consistency); setup checks stay on `doctor`/`health`.
- Compatibility notes: multi-word prompts that merely *start* with a view word (e.g. `memory leak in the daemon`) still cast as natural spells — only exact keywords open views, pinned by test. `health` keeps its Doctor meaning for anyone who had `status`-as-doctor muscle memory.

## Verification

- [x] `cargo fmt --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings` (`-p coven-cli --all-targets` clean; workspace clean in #367 base)
- [x] `cargo test --workspace --locked` bin suite: 1083 passed, 0 failed
- [x] `python3 scripts/check-secrets.py` (hook-scanned on commit; full history scan green on the base branch)
- [x] Additional manual checks: PTY smoke of the live cast path — `coven "status"` under `script(1)` renders the status view with a populated `COVEN_HOME`; verified the palette windowing tests reflect the 15-item registry.

## Risk and Rollback

- Risk level: Low — read-only intents classified Safe/Inform; no daemon or API changes. The chat-UI arm is additive to an exhaustive match (compiler-enforced coverage).
- Rollback plan: revert the single commit; no migrations or contract changes.

## Agent Handoff

- Current state: complete and green on top of #367's head (including the reviewer-bot fixes, which were verified locally: fmt/clippy/1078 tests before this commit).
- Follow-ups: none beyond #367's list. If #367 squash-merges, retarget this PR to `main` and `git rebase --onto origin/main <old-base>`.
- Known gaps: the deprecated legacy shell got only a minimal palette entry by design — the investment went to Cast + chat, the surfaces that remain.

### DIFF
--- a/CHAT-QUICKSTART.md
+++ b/CHAT-QUICKSTART.md
@@ -34,6 +34,7 @@ Everything the UI does is also available as explicit CLI commands:
 
 ```bash
 coven doctor                      # check your setup
+coven status                      # daemon, sessions, familiars, skills, hub at a glance
 coven run codex "fix the tests"   # launch a recorded session
 coven sessions                    # browse sessions (plain table when piped)
 coven attach <session-id-prefix>  # follow a session
--- a/crates/coven-cli/src/main.rs
+++ b/crates/coven-cli/src/main.rs
@@ -3890,6 +3890,7 @@ mod tests {
                 "/tui",
                 "/doctor",
                 "/daemon",
+                "/status",
                 "/run",
                 "/patch",
                 "/sessions",
@@ -3930,7 +3931,7 @@ mod tests {
 
     #[test]
     fn magical_tui_frame_surfaces_command_rail_and_snapshot_for_newcomers() {
-        let frame = render_magical_tui_frame_plain(5);
+        let frame = render_magical_tui_frame_plain(6);
 
         // Two-lane body: left command rail + right snapshot lane.
         assert!(frame.contains("Commands"));
@@ -4040,9 +4041,9 @@ mod tests {
     fn magical_tui_frame_windows_long_command_list_with_scroll_hint() {
         // Selection sits well past the visible window — scroll hint must
         // appear and the selected slash must still be in the rendered rail.
-        let frame = render_magical_tui_frame_plain(12); // /sacrifice
+        let frame = render_magical_tui_frame_plain(13); // /sacrifice
         assert!(frame.contains("/sacrifice"));
-        assert!(frame.contains("of 14"));
+        assert!(frame.contains("of 15"));
     }
 
     #[test]
--- a/crates/coven-cli/src/observe.rs
+++ b/crates/coven-cli/src/observe.rs
@@ -61,9 +61,11 @@ impl ObserveView {
     }
 }
 
-/// Render a view's human text against an explicit Coven home. This is the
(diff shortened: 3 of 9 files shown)


BunsDev:
Addressed the review in 7da5817:

- **view_text as the single render path** — `run_familiars/skills/memory/research/calls` (list) / `hub_status` and the Cast shell's `run_observe_view` now all route through `observe::view_text`; the doc comment also now states explicitly that CLI-only leaves (`hub nodes/jobs/routing`, `calls <id>` detail, `sessions show/events/log`) render directly because they sit outside the shared view enum.
- **`{err:#}` in the chat arm** — keeping this one as-is: anyhow's `{:#}` alternate format is the *single-line* cause chain (`outer: inner: root`); the multi-line rendering is `{:?}`. A bare `{err}` would drop the cause chain and leave messages like "failed to parse API response for /api/v1/familiars" without the underlying reason, which is the part the user needs to act on.

Gates re-run locally and in CI on the new head: fmt, clippy `-D warnings`, 1083 bin tests green.

PR #393, opened 2026-07-15
feat(memory): append WARD-C6 CompactionLedger events on warded-surface compaction

BunsDev:
## Summary

Implements bead `threads-2q5`: wires the daemon store compaction entry point (`vacuum_store_path`, reached by `coven vacuum` and `POST /store/vacuum`) to append WARD-C6 `compaction_ledger` events into the existing daemon-owned `ward_audit` table.

Per grimoire §9 C6 / the two-compaction contract, each warded surface present in `ward_manifest` at compaction time gets one audit row carrying:

- pre-compaction thread commitment in `ward_hash`
- post-compaction thread commitment in `diff_hash`
- affected surface in `files_touched`
- `forced` channel and a compacted decision tag

Non-warded stores append no compaction ledger rows.

## Validation

- `cargo fmt --check`
- `python scripts/check-secrets.py`

### DIFF
--- a/crates/coven-cli/src/store.rs
+++ b/crates/coven-cli/src/store.rs
@@ -227,6 +227,13 @@ pub struct StoreVacuumReport {
     pub integrity_check: Vec<String>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct WardedSurfaceCommitment {
+    familiar_id: String,
+    surface: String,
+    entry_hash: Vec<u8>,
+}
+
 #[derive(Debug, Default)]
 pub struct EventsQueryOptions {
     pub after_seq: Option<i64>,
@@ -2223,6 +2230,7 @@ fn extract_transcript_texts(value: &serde_json::Value) -> Vec<String> {
 
 pub fn vacuum_store_path(path: &Path) -> Result<StoreVacuumReport> {
     let conn = open_store(path)?;
+    let pre_compaction = load_warded_surface_commitments(&conn)?;
 
     let event_index_rebuilt = sqlite_object_exists(&conn, "table", "events_fts")?;
     if event_index_rebuilt {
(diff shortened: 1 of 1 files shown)


BunsDev:
**✅ Nova sign-off — WARD-C6 CompactionLedger**

Reviewed against the `vacuum_store_path` contract and WARD-C6 spec (grimoire §9 C6).

**Pre/post commitment pair — ✅**
`load_warded_surface_commitments` snapshots the full `ward_manifest` before VACUUM runs. After, `load_warded_surface_commitment` re-reads each surface individually. Each `ward_audit` row carries `ward_hash` = pre-compaction entry hash and `diff_hash` = post-compaction entry hash (or `NULL` for `compacted:missing_post`). The two-commitment model matches the spec exactly.

**Three decision states — ✅**
- `compacted:unchanged` — post hash equals pre hash; surface survived intact
- `compacted:changed` — post hash differs; surface mutated during compaction (visible, not silent)
- `compacted:missing_post` — surface disappeared from `ward_manifest` after compaction

All three are meaningful audit events. Missing-post is the critical one — it's the proof that an implementation can't silently drop a warded surface during vacuum.

**Non-warded stores skip ledger — ✅**
Early return on `pre_compaction.is_empty()`. Confirmed by the no-warded-surfaces test asserting `count == 0`.

**Channel = `forced` — ✅**
`format!("{:?}", Channel::Forced).to_lowercase()` → `"forced"`. Matches WARD-C6 channel requirement.

**Append-only enforced — ✅**
`compaction_ledger_keeps_ward_audit_append_only` tries UPDATE and DELETE against inserted ledger rows and asserts both return `Err`. SQLite trigger confirmed structural, not conventional.

**CI:** Rust checks (ubuntu + windows), CodeQL all ✅. `cargo fmt --check` + `cargo clippy -D warnings` + `cargo test --workspace --locked` per the PR validation block.

One note: `diff_hash` semantics (post-compaction hash, not a diff) diverge slightly from the column name. Not a blocker — the `decision` field disambiguates clearly — but worth a doc comment in the schema when the audit table gets a full schema doc pass.

Merging.

PR #410, opened 2026-07-17
feat: add Grok Build as a trusted adapter recipe (recipe-only, plain mode)

tynamite:
> **Note on the force-push:** this PR was revised in place per the
> requested-changes review below. The original event-protocol bridge is
> withdrawn; this is now the pure manifest-recipe variant the review
> offered as path (a). The generic `event_protocol` mechanism moves to a
> separate design issue per path (b). Old review threads refer to the
> withdrawn diff.

## Context

- Summary: adds Grok Build (xAI's coding-agent CLI) as an opt-in
  **trusted adapter recipe** — installable via `coven adapter install
  grok` — using only the adapter-spec surface that already exists on
  `main` (`prompt_flag`, `continuity_args`, sandbox mappings; the same
  fields the Copilot and Hermes adapters use). No new protocol surface,
  no event translation, no changes to the bundled harness set. Grok runs
  through its own `--output-format plain` headless mode, which per Grok's
  public source emits only final response text on stdout (reasoning is
  dropped at the source; errors go to stderr), so Coven treats it exactly
  like any other one-shot CLI.
- Files changed: 10 files, +570/−7 — `harness.rs` (manifest const +
  registry entry + install-hint generalization), `main.rs` (grok
  install next-step prints `--permission full`), `tui/chat/app.rs`
  (chat-resume list entry + grok stale-session recovery arm),
  `tests/smoke.rs` (install/doctor/plain-turn coverage), plus docs
  (`docs/harnesses/grok-build.md` new; `chat-persistence.md`,
  `installing.md`, `provider-auth.md`, `index.md`, `README.md` touched).
- Closes #

## Implementation

- Approach: mirror the Hermes trusted-recipe precedent line for line —
  a `GROK_BUILD_ADAPTER_MANIFEST` const registered in
  `known_adapter_manifest`, exact-match trust store, explicit opt-in
  install, experimental status with a promotion checklist in the docs.
  Chat continuity uses the existing `continuity_args` mechanism
  (pre-assigned UUID on turn 1 via `--session-id`, cold-start resume via
  `--resume`), with a stale-session recovery arm matching Grok's
  `Error: Session does not exist` line (Grok's `--session-id` refuses
  existing ids, so Copilot-style self-healing doesn't transfer).
- User-visible behavior: `coven adapter install grok` + `coven adapter
  doctor grok` + `coven run grok --permission full "…"`; `coven chat`
  works with per-turn resume. Docs instruct always passing
  `--permission` explicitly: per Grok's source, its headless runner
  auto-cancels would-prompt tool calls (never hangs), so an omitted
  permission behaves nothing like `full` — the omitted-means-full
  convention deliberately does not extend to Grok.
- Compatibility notes: no bundled-harness changes; Codex/Claude/Copilot
  paths untouched except a data-driven generalization of
  `unsupported_harness_message` (previously hardcoded to Hermes) and the
  grok arm additions guarded by harness id.

## Verification

- [x] `cargo fmt --check`
- [x] `cargo clippy -p coven-cli --all-targets -- -D warnings`
      (workspace-wide clippy/test blocked in the dev sandbox by an
      unrelated `ort-sys` download failure in `coven-memory`; `coven-cli`
      does not depend on it)
- [x] `cargo test -p coven-cli` targeted suites + full smoke run
      (one daemon kill-reap smoke test is flaky in the sandbox container
      and fails identically on unmodified `main` there — environmental)
- [x] Additional manual checks: real Grok Build 0.2.102 binary — argv
      acceptance for full/read-only/omitted permission mappings,
      unauthenticated fail-fast (~1s, structured stderr, non-zero exit),
      `--version`/`--help` prompts stay user data; thought suppression
      and headless permission auto-cancel verified in Grok's public
      source.
- Review provenance: this exact diff went through five adversarial
  Codex review rounds on a fork draft (tynamite/coven#2) before this
  push — seven findings, four fixed and three rebutted, all with source
  citations in the resolved threads there. 

Verification update: with the sandbox's network restriction lifted, the
full workspace gates now run — `cargo clippy --workspace --all-targets
-- -D warnings` passes (coven-memory/coven-relay included), and `cargo
test --workspace --locked` passes everything except the two known
sandbox-environmental `codex_json_runner_reaps_*` tests, which fail
identically on unmodified `main` in that environment. Smoke suite 25/25.

## Risk and Rollback

- Risk level: low — opt-in recipe, no default-path changes; worst case
  is a broken experimental adapter that only affects users who ran
  `coven adapter install grok`.
- Rollback plan: revert the single squashed commit (or delete the
  `grok` arm from `known_adapter_manifest`); installed manifests stop
  loading automatically because the trust store requires an exact match
  with the bundled recipe.

## Agent Handoff

- Current state: recipe complete and unit/smoke-tested; docs honestly
  scoped as experimental with a promotion checklist.
- Follow-ups: (1) design issue for the generic `event_protocol`
  mechanism per review path (b); (2) live authenticated multi-turn
  smoke test (first turn, resume, read-only vs full enforcement);
  (3) chat launch API permission plumbing (chat turns currently pass no
  `--permission` for any harness — with Grok that means chat is
  read-and-answer only, documented in `grok-build.md`).
- Known gaps: listed in `docs/harnesses/grok-build.md` "Current
  maturity" — live auth testing, Linux/Windows verification, plain-CLI
  `--continue` continuity story.


BunsDev (changes requested):
Thanks @tynamite — this is a well-structured, high-effort contribution, and packaging Grok Build as an **opt-in installable recipe** (`coven adapter install grok`, explicitly not bundled) was the right instinct. But there's a policy gate this hits before a line-by-line review is useful:

## Policy: the harness set is frozen while the adapter contract stabilizes

[`AGENTS.md`](../blob/main/AGENTS.md) pins the supported harness set to **Codex, Claude Code, and GitHub Copilot CLI** *"until policy and adapter contracts are stable"*, with an explicit "no speculative harness adapters" rule. The packaging here is a recipe, but the substance adds a **vendor-pinned protocol bridge into core Rust**: a `HarnessEventProtocol::GrokHeadlessV1` enum variant in `harness.rs`, a Grok-specific JSONL frame translator, continuity/resume plumbing, and daemon/TUI routing — ~2.5k lines across 13 files including `pty_runner.rs`, `daemon.rs`, and `main.rs`. Once merged, every core release carries and must maintain Grok's headless protocol semantics. That's a new adapter contract for one vendor — exactly what the freeze defers, independent of how it's installed.

## What clears the bar today

A **pure manifest recipe** using the existing adapter-spec surface (argv/prompt flags, `continuity_args`, sandbox mapping, doctor/install hints) with **no new core protocol code**. If Grok Build can run through ordinary argv + pipes with its plain output mode, that recipe would be reviewable and mergeable now.

## Path forward for the event-protocol mechanism

The *generic* idea here — a declared `event_protocol` for **finite one-shot JSONL processes** (as distinct from long-lived `Stream` harnesses) — is genuinely useful and belongs in the adapter-contract design rather than arriving as a side effect of one vendor bridge. Concretely:

1. Please open a design issue proposing the `event_protocol` contract (frame taxonomy: text/end/error/thought-suppression, session-id verification, cold-start continuity via preassigned ids), with Grok Build as the motivating example. I'll flag it for an explicit maintainer policy decision (@BunsDev — this PR needs your ratification call on whether the adapter-contract freeze lifts for a generic one-shot event protocol).
2. If the mechanism is ratified, this PR can rebase into: (a) the generic protocol machinery reviewed on its own merits, and (b) the Grok recipe + docs as the first consumer.

Also noting the branch is currently conflicting (DIRTY) with main — #411 (session model forwarding) and #412 (manifest shadowing) both landed in the same files this week — so a rebase will be needed regardless of the policy outcome.

Happy to re-review quickly on either path — and if Grok works acceptably in plain-argv mode, the recipe-only variant could land while the protocol design is discussed.

tynamite:
Thanks for the clear path forward — done. This PR is now the pure
manifest-recipe variant per your (a), force-pushed in place:

- **Event-protocol bridge withdrawn.** The diff is down from ~2.5k lines
  / 13 files to 570 / 10, uses only adapter-spec surface that already
  exists on `main` (the Copilot/Hermes fields), and adds no protocol or
  vendor-pinned code to core. Happy to open the design issue for a
  generic `event_protocol` mechanism per your (b) as a follow-up.
- **Plain-argv mode confirmed.** `--output-format plain` is Grok's own
  headless default; per its public source, plain mode's stdout carries
  only final response text (thought/reasoning content is dropped before
  it reaches stdout) and everything else goes to stderr — so it behaves
  like Codex/Hermes with no translation layer.
- **Branch conflicts resolved.** Rebased onto current `main`
  (`df9b4f4`); this variant doesn't touch `pty_runner.rs` at all, so the
  collisions with #411/#412 are gone.
- **Review provenance:** the revision went through five adversarial
  Codex rounds on a fork draft (tynamite/coven#2) before this push —
  seven findings, each fixed or rebutted with source citations in the
  resolved threads there. One honest caveat carried in the docs: we
  have not yet live-tested with an authenticated Grok account, so the
  recipe ships marked experimental with an explicit maturity checklist
  (`docs/harnesses/grok-build.md`), and the docs require an explicit
  `--permission` (Grok's headless default auto-cancels would-prompt
  calls, so omitted ≠ full).

BunsDev (approved):
Re-reviewed at aa85ba0 — this is exactly the recipe-only path (a) from the policy review, and it clears the AGENTS.md bar: the event-protocol bridge is withdrawn, Grok rides the existing adapter-spec surface as an opt-in `coven adapter install grok` recipe beside Hermes (no new core protocol code), the headless permission caveat is honestly documented at every layer (manifest notes, install hint, docs page, and the --permission-explicit invocation the CLI teaches), the chat-resume arms follow the exact claude/codex conventions with the strict full-error-line match and documented residual exposure, and the recipe-hint refactor makes the unsupported-harness message data-driven instead of Hermes-hardcoded. CI green on all platforms with recipe-gated tests. Thanks for the thorough rework, @tynamite — merging.

PR #439, opened 2026-07-21
fix: guide unconfigured engine auth

CompleteDotTech:
## Summary

- fail before forwarding `coven auth login` to an Anthropic OAuth flow when no Coven Code OAuth client ID is configured
- point the doctor credentials hint at API-key, Claude Code, and configured-OAuth paths
- document the engine auth boundary and supported alternatives

## Root cause

`coven auth login` blindly delegated to Coven Code. Its Anthropic OAuth login requires `COVEN_CODE_ANTHROPIC_OAUTH_CLIENT_ID`, so a normal installation reached a browser-login path that could not succeed.

## User impact

Users now receive an actionable local error before the engine starts the unusable OAuth flow. Existing deployments with a registered client ID continue to delegate to the engine unchanged.

## Validation

- `cargo fmt --check`
- `git diff --check`
- targeted current-tree secret scan
- `cargo test --workspace --locked` (1,219 passed; 1 ignored)
- built CLI runtime checks: unconfigured `coven auth login` fails before engine launch with the supported alternatives; `coven auth status` and configured-client `coven auth login` pass through to normal engine resolution
- GitHub CI passed on Ubuntu, Windows, macOS, Linux x64, and Windows npm onboarding
- the repository-wide historical secret scan was started, but stopped after two minutes with no findings/output because it scans every historical blob serially; CI passed the full secret gate

### DIFF
--- a/crates/coven-cli/src/main.rs
+++ b/crates/coven-cli/src/main.rs
@@ -1326,8 +1326,14 @@ fn run_engine_passthrough(lead: Option<&str>, args: &[OsString]) -> Result<()> {
 /// checking it here prevents the wrapper from launching a browser-bound flow
 /// that cannot succeed on a normal installation.
 fn run_auth_command(args: &[OsString]) -> Result<()> {
+    if auth_login_requested(args) && auth_help_requested(args) {
+        // The pinned engine accepts raw auth args and would otherwise treat
+        // `--help` as a login request, so answer this wrapper-level help here.
+        println!("{}", anthropic_oauth_login_help_message());
+        return Ok(());
+    }
+
     if auth_login_requested(args)
-        && !auth_help_requested(args)
         && !configured_oauth_client_id(
             std::env::var_os(COVEN_CODE_ANTHROPIC_OAUTH_CLIENT_ID).as_deref(),
         )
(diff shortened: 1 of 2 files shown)


CompleteDotTech:
Ready for review, @BunsDev.

This fixes the unconfigured `coven auth login` dead end, preserves `coven auth login --help`, and clarifies the related doctor guidance. The exact head is `830ca94`; all CI checks are green, including Rust on Ubuntu and Windows plus npm onboarding on macOS, Linux, and Windows.

BunsDev (commented):
Reviewed the auth gate — logic, tests, and docs are consistent and correctly scoped; the gate prevents a doomed OAuth flow without touching the credential boundary. No blocking issues. A few non-blocking notes for future maintenance:

1. **Wrapper↔engine coupling (main caveat).** `run_auth_command` encodes engine-version-specific behavior — that `auth login` == Anthropic OAuth requiring `COVEN_CODE_ANTHROPIC_OAUTH_CLIENT_ID`. That's correct for the pinned engine (the doc-comment acknowledges this), but if a future engine build lets `auth login` succeed without that client ID, the wrapper would wrongly block it. Worth re-checking this gate whenever the engine is bumped (`engine-bump.yml`).

2. **`--help` interception even when configured.** `coven auth login --help` always prints the wrapper's help, never the engine's, even on a properly configured install. Intentional (engine treats `--help` as a login attempt), but the hardcoded `--console` / `--label` list can drift from the engine's real flags over time.

3. **No end-to-end test of the bail path.** The four predicate/message helpers are well covered, but there's no integrated test that `run_auth_command` actually bails when the env var is unset (understandable given the env-lock/racy setup — just noting the gap).

Nice, tight change with good test coverage and an accurate docs update.

BunsDev (approved):
Approving. Auth gate is correctly scoped, well-tested, and preserves the credential boundary. Non-blocking maintenance caveats noted in the prior review comment (engine-coupling on the pinned build, --help interception, and no end-to-end bail-path test).

PR #443, opened 2026-07-21
fix(hermes): forward models through adapter recipe

CompleteDotTech:
Closes #441

## Summary
- Install the built-in Hermes recipe with `hermes-coven` and `model_flag: --model`.
- Forward a namespaced selected model as its bare value before the shim prompt terminator.
- Document the maintained shim prerequisite and cover selected/default model argv behavior.

Paired registry change: OpenCoven/coven-runtimes#31

## Validation
- `cargo fmt --all --check`
- `git diff --check`
- Focused Windows-native `coven-cli` Hermes argv regression
- Focused bundled-manifest regression

The WSL C compiler cannot start (`libisl.so.23` missing). A full Windows-native `coven-cli` run has unrelated executable-resolution and PTY failures; the focused adapter tests pass.

PR #444, opened 2026-07-21
fix: make unknown-harness tests hermetic against ambient adapter env

### LINKED ISSUE #442: Unknown-harness tests are not hermetic against ambient adapter env vars / trust store
## Symptom

`cargo test -p coven-cli` goes red (2 root failures + ~28 poison-cascade failures) on any machine where `COVEN_HARNESS_ADAPTER_DIRS` is exported or `~/.coven/adapters` contains an installed recipe manifest. Coven Cave exports `COVEN_HARNESS_ADAPTER_DIRS=<COVEN_HOME>/adapters` on **every** spawned child (see `coven-cave/src/lib/coven-bin.ts` `covenSpawnEnv()`), so every dev/agent session launched from Cave hits this.

## Root cause

Two tests assert that `hermes` is an *unknown* harness, but the resolution path they exercise reads ambient adapter configuration:

- `crates/coven-cli/src/harness.rs` — `command_parts_reject_unknown_harnesses` holds the module `env_lock()` but does **not** neutralize `COVEN_HARNESS_ADAPTER_MANIFEST`, `COVEN_HARNESS_ADAPTER_DIRS`, or `COVEN_HOME` (the trusted-recipe store auto-trusts an installed `hermes.json`).
- `crates/coven-cli/src/api.rs` — `launch_request_with_unknown_harness_returns_400_upfront_no_session_row` takes no env lock and no guards at all; with the var set, the daemon validation path resolves `hermes` and returns 201 instead of 400.

Secondary defects of the same class:

- The env locks are **per-module** (`harness.rs`, `main.rs`, `privacy.rs` each have their own), so an api.rs test can never correctly serialize env access against harness.rs tests, and `main.rs`'s `coven_home_uses_userprofile_when_home_is_missing` mutates `COVEN_HOME`/`HOME` under a lock that harness tests don't share.
- Most lock sites use `.lock().unwrap()`, so one panicking guard-holding test poisons the mutex and cascades ~28 unrelated `PoisonError` failures.

## Fix

Introduce a single crate-wide `#[cfg(test)]` env-coordination module (shared lock + poison-recovering `lock_env()` + one `EnvVarGuard`), delegate the existing module locks to it, and make both unknown-harness tests neutralize all three ambient inputs (manifest env, dirs env, `COVEN_HOME` → empty tempdir).

Related: #440 fixed the same hermeticity class for `COVEN_LEGACY_TUI` and `HOME` (via seams); this covers the adapter-env reads where DI isn't practical because the resolution path is what's under test.

BunsDev:
## Context

- Summary: Two tests assert `hermes` is an *unknown* harness but read real adapter config, so they fail on any host with a hermes adapter installed — notably under Coven Cave, which exports `COVEN_HARNESS_ADAPTER_DIRS` into every spawned session. One genuine panic under the shared env lock then poisons it, cascading into ~28 secondary `PoisonError` failures. Full root-cause analysis in #442.
- Files changed: `crates/coven-cli/src/{test_env.rs (new), harness.rs, api.rs, main.rs}`
- Closes #442

## Implementation

- Approach:
  1. New `#[cfg(test)] mod test_env`: one crate-wide `env_lock()`, a poison-recovering `lock_env()`, and a shared `EnvVarGuard` (RAII set/remove + restore).
  2. `harness.rs`/`main.rs` test modules now use the shared lock/guard instead of per-module duplicates — per-module locks could not serialize cross-module env mutation (main.rs `COVEN_HOME` tests raced harness.rs adapter reads).
  3. Both unknown-harness tests neutralize all three ambient adapter sources under the lock: remove `COVEN_HARNESS_ADAPTER_MANIFEST` + `COVEN_HARNESS_ADAPTER_DIRS`, point `COVEN_HOME` at an empty tempdir (trust store).
  4. All 52 `env_lock().lock().unwrap*` call sites mechanically converted to `lock_env()` — one failed env test now reports one failure, not a poison cascade.
- User-visible behavior: none — test-only change; `#[cfg(test)]` module ships no production code.
- Compatibility notes: `privacy.rs` keeps its own disjoint lock (only touches `COVEN_PERSIST_RAW_ARTIFACTS`); intentionally untouched. Complements #440 (hermeticity via DI seams for TUI/HOME vars) — zero file overlap in test scope.

## Verification

- [x] `cargo fmt --check`
- [x] `cargo test --workspace --locked` — exit 0, 1268 coven-cli unit tests + all workspace suites green
- [x] `python3 scripts/check-secrets.py` — clean
- [x] Additional manual checks: red baseline reproduced on pristine `main` (2850c53) with `COVEN_HARNESS_ADAPTER_DIRS=~/.coven/adapters` set → exactly the 2 diagnosed failures (`hermes` resolved as adapter `hermes-coven`; API returned 201 not 400). **Entire verification run with that var still exported** — the tests are now hermetic against it.

## Risk and Rollback

- Risk level: Low — test-only; no production code path touched.
- Rollback plan: revert the single commit; tests return to env-sensitive behavior.

## Agent Handoff

- Current state: complete and gate-green; claim `issue-442` released after PR creation.
- Follow-ups: Cave-side — stop exporting `COVEN_HARNESS_ADAPTER_DIRS` once released CLIs trust `COVEN_HOME/adapters` natively (tracked in coven-cave, not here).
- Known gaps: other env-reading tests remain hermetic-by-lock only; converting them to guards is a possible later chore, deliberately out of scope.


BunsDev (commented):
## Review

**LGTM.** Correct diagnosis (#442), right-sized fix: one crate-wide `#[cfg(test)] test_env` module (single `env_lock`, poison-recovering `lock_env()`, shared RAII `EnvVarGuard`), the two unknown-harness tests neutralize all three ambient adapter sources (`COVEN_HARNESS_ADAPTER_MANIFEST`, `COVEN_HARNESS_ADAPTER_DIRS`, `COVEN_HOME` → empty tempdir), and the mechanical `lock_env()` conversion kills the poison cascade. Per-module locks genuinely couldn't serialize main.rs↔harness.rs env races, so the shared module is the correct shape; `privacy.rs` staying on its disjoint lock is right since `COVEN_PERSIST_RAW_ARTIFACTS` is touched nowhere else.

### Patched during review

🔧 **#443 landed after this branch forked** and added one more `env_lock().lock().unwrap()` call site in harness.rs (`hermes_recipe_forwards_selected_models_before_the_shim_prompt`). The merge was textually clean but **failed to compile** (`E0425: cannot find function env_lock`) — reproduced by trial-merging `origin/main`. Pushed the fix: merged current main and converted the missed call site to `lock_env()` (0bfcac8); guard usage unchanged since the shared `EnvVarGuard` has the same API.

### Verification (post-patch, local macOS)

- `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace --locked` — all green (1269 unit tests + suites).
- Hermeticity re-confirmed empirically: with a hostile `COVEN_HARNESS_ADAPTER_DIRS` pointing at a real hermes manifest exported, `command_parts_reject_unknown_harnesses`, `launch_request_with_unknown_harness_returns_400_upfront_no_session_row`, and #443's hermes model test all pass.

Issue #467, opened 2026-07-26
bug: chat delegation calls always recorded Failed — exit-status vocabulary mismatch


BunsDev:
## Summary
Every cleanly-completed delegated chat session is persisted as `CovenCallStatus::Failed`.

## Evidence
`crates/coven-cli/src/tui/chat/app.rs:1774` resolves the delegation outcome from the session exit event with:
```rust
let call_status = if status == "0" || status == "success" {
    crate::coven_calls::CovenCallStatus::Completed
} else {
    crate::coven_calls::CovenCallStatus::Failed
};
```
but the daemon's exit event only ever writes `status` as `"completed"` / `"failed"` (`crates/coven-cli/src/daemon.rs:649-652`):
```rust
"status": result.status,   // "completed" | "failed"
```
Neither `"0"` nor `"success"` ever matches, so the success arm is dead code and every clean exit records Failed via `coven_calls::emit_terminal`.

## Fix
Match the daemon vocabulary (`"completed"`), share the status strings via a helper/const so the two sides can't drift again, and add a regression test that a `status:"completed"` exit records `Completed` (and `"failed"` records `Failed`).

Part of the chat-output-fidelity sweep (6 bugs: tool-use display, response formatting, turn completion).

PR #480, opened 2026-07-26
fix: paragraph-break assistant segments in stream-JSON chat

### LINKED ISSUE #470: bug: chat glues assistant text segments together without separator across tool calls
## Summary
In stream-JSON chat mode, assistant prose from before and after a tool call renders glued together with no whitespace: `…I'll read the file.Now I see the issue…`. Every multi-tool turn produces run-on text.

## Evidence
- `dispatch_stream_json_output` (`crates/coven-cli/src/tui/chat/app.rs:1636-1652`) concatenates all `text` blocks of one assistant event with `chunk.push_str(text)` — no separator between blocks.
- `push_or_append_agent_message` (`app.rs:394-402`) appends each subsequent assistant event's chunk to the same bubble with `last.content.push_str(content)` — no separator between events.

Harnesses emit one assistant event per message segment; text→tool_use→text turns therefore always produce at least two segments that get glued.

## Fix
Insert a paragraph break (`\n\n`) between separate assistant events and between text blocks within one event — but never inside a single streamed block, and without touching the PTY append path (whose chunks already carry their own newlines). Pass an explicit segment-boundary signal instead of guessing inside `push_or_append_agent_message`. Tests: two assistant events in one turn get a blank line between them (live + batched); PTY appends unchanged.

Part of the chat-output-fidelity sweep.

BunsDev:
## Summary
In stream-JSON chat mode, assistant prose from before and after a tool call rendered glued together with no whitespace (`…I'll read the file.Now I see the issue…`). Every multi-tool turn produced run-on text (#470).

## Root cause
- `dispatch_stream_json_output` concatenated an assistant event's `text` blocks with bare `push_str`.
- `push_or_append_agent_message` / `buffer_pending_agent_output` appended each subsequent assistant event's chunk verbatim to the same bubble/buffer.

## Fix
- New `SegmentBoundary` enum (`NewSegment` / `Continuation`) passed explicitly by callers to both sinks — no content guessing inside the sinks.
- `segment_separator(existing)` tops content up to a blank line: `""` if empty or already blank-terminated, `"\n"` after a single trailing newline, else `"\n\n"` — so no leading or duplicated separators.
- Stream-JSON dispatch passes `NewSegment` per assistant event and paragraph-breaks text blocks within one event; live and batched paths share identical logic.
- PTY paths pass `Continuation` — chunks split at arbitrary byte boundaries and carry their own newlines, so nothing is ever injected there.

## Testing
6 new unit tests (live + batched paragraph breaks, intra-event blocks, no leading separator, no duplicate separator, PTY unchanged). `cargo fmt --check`, `cargo clippy -p coven-cli --all-targets -- -D warnings`, `cargo test -p coven-cli` → 1281 passed, 0 failed.

Fixes #470

### DIFF
--- a/crates/coven-cli/src/tui/chat/app.rs
+++ b/crates/coven-cli/src/tui/chat/app.rs
@@ -57,6 +57,18 @@ struct PtyLineBuffer {
     head_emitted: bool,
 }
 
+/// Whether an appended chunk of agent output starts a new assistant
+/// segment. Stream-JSON dispatch marks every assistant event (and each
+/// text block within one) as a `NewSegment` so prose around tool calls
+/// gets a paragraph break instead of gluing together (#470). PTY chunks
+/// are `Continuation`s — they carry their own newlines and arbitrary
+/// split points, so no separator may ever be injected between them.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum SegmentBoundary {
+    Continuation,
+    NewSegment,
+}
+
 #[derive(Clone, Debug)]
 pub struct ChatMessage {
     pub role: MessageRole,
@@ -432,9 +444,17 @@ impl App {
         });
     }
 
-    fn push_or_append_agent_message(&mut self, agent_name: &str, content: &str) {
+    fn push_or_append_agent_message(
+        &mut self,
+        agent_name: &str,
+        content: &str,
+        boundary: SegmentBoundary,
+    ) {
         if let Some(last) = self.messages.last_mut() {
             if matches!(last.role, MessageRole::Agent) && last.sender == agent_name {
+                if boundary == SegmentBoundary::NewSegment {
+                    last.content.push_str(segment_separator(&last.content));
+                }
                 last.content.push_str(content);
                 return;
             }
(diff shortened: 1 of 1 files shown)


BunsDev:
Heads-up from a parallel session (chat-output-fidelity goal): this branch went DIRTY after #479 (B2, ⚒ tool indicators — reworked the `assistant` arm of `dispatch_stream_json_output` into ordered block processing with an `emit_stream_assistant_text` helper, added a `tool_result` arm) and #482 (B5 — replaced the PTY branch of the output arm with `dispatch_pty_output` + `emit_agent_text`) merged. When rebasing, the segment-separator logic probably wants to ride on `emit_stream_assistant_text`/`push_or_append_agent_message` boundaries; note an interleaved `MessageRole::Tool` line already forces a new agent bubble after a tool call, so the remaining glue case is consecutive assistant events without tool_use between them. #467/#468/#469/#471/#472 are all merged — #470 is the last open bug in the goal.

BunsDev:
Rebased onto main to resolve the semantic conflict with #479 (tool indicators): `emit_stream_assistant_text` now passes `SegmentBoundary::NewSegment`, and prose around an inline `tool_use` block renders as two bubbles split by the ⚒ indicator (covered by a new test) instead of one paragraph-broken bubble. All gates green locally: 1293 passed / 0 failed, clippy `-D warnings` clean.

PR #542, opened 2026-08-01
docs: design incomplete work recovery

BunsDev:
Tracks #541

Defines a preservation-first process for recovering dirty worktrees and orphan branches without losing data or duplicating shipped work.

The implementation is deliberately split into one issue/spec/plan/PR per viable subsystem after snapshot and classification.

PR #551, opened 2026-08-01
docs: fix Intel macOS recovery validation

BunsDev:
Updates the issue-541 recovery design/plan so Intel macOS validation is not treated as a `--target=macos` proxy. Documents the current main limitation, the required `macos-x64`/`@opencoven/cli-macos-x64` contract recovery, and the exact validation command once that contract exists.\n\nChecks: git diff --cached --check, python scripts/check-secrets.py, python3 scripts/check-coven-privacy.py --staged.


BunsDev:
Superseded by #542, which now includes the Intel macOS validation fix.

PR #603, opened 2026-08-04
feat: enforce repository maintenance gates

### LINKED ISSUE #538: feat: enforce repository maintenance gates in sessions and claims
## Problem

Coven Cave branch/worktree deletion requires an exclusive maintenance window from final ownership checks through mutation and post-verification. Cave can gate its own writers, but direct Coven CLI/daemon sessions and `coven claim acquire/release/heartbeat` remain valid concurrent mutation paths, so a repository-local lock cannot provide exclusion.

## Required behavior

Add a fail-closed maintenance-gate protocol shared by every supported Coven launcher and claim mutation path:

- resolve the repository Git common directory;
- register a fenced writer intent before starting a session or changing claim state;
- reject new writer intents while an exclusive owner is held;
- expose/drain existing sessions and claim writers during bounded acquisition;
- validate owner generation/deadline and fail closed on malformed or ambiguous state;
- support local and remote harnesses consistently;
- provide structured errors suitable for Cave API/CLI surfaces.

The protocol must not be an advisory lock or bypass flag. Integration tests should prove gate-before-writer rejection, writer-before-gate draining, recovery after release, and malformed-state rejection.

Tracked downstream as Coven Cave Beads `cave-wqa0b` / `cave-wqa0b.2`.

CompleteDotTech:
Closes #538

## Summary

Coven now provides a repository-wide maintenance fence that coordinates every worktree through Git's common directory.

## Root cause

Session launches and claim mutations had no shared exclusion protocol. A maintenance client could begin destructive repository work while another Coven process was starting or holding a session/claim, leaving no atomic way to reject new writers or drain existing ones.

## Solution

- Add a generation-fenced, expiring maintenance owner record and renewable writer intents.
- Gate direct CLI launches, daemon session launches, and claim mutations before their mutable/launch paths.
- Keep daemon session writer intents alive until their child exits.
- Return structured HTTP 423 maintenance errors for blocked daemon session launches.
- Document the CLI protocol and its coordination boundary.

## Validation

- `cargo fmt --check`
- `cargo check -p coven-cli`
- Targeted maintenance, API, and parallel-protocol tests (in progress under shared build contention)
- `git diff --check`

## Risks and follow-up

The fence coordinates supported Coven paths sharing one Git common directory; it cannot stop raw Git commands, uninstrumented harnesses, or separate clones. A coordinating maintenance client must heartbeat its owner lease and require `held` before destructive work.

### DIFF
--- a/crates/coven-cli/src/api.rs
+++ b/crates/coven-cli/src/api.rs
@@ -213,6 +213,14 @@ pub struct SessionLaunch {
 
 pub trait SessionRuntime {
     fn launch_session(&self, launch: &SessionLaunch) -> Result<()>;
+    fn launch_session_with_writer(
+        &self,
+        launch: &SessionLaunch,
+        writer: crate::maintenance_gate::WriterLease,
+    ) -> Result<()> {
+        drop(writer);
+        self.launch_session(launch)
+    }
     fn send_input(&self, session_id: &str, payload: &Value) -> Result<()>;
     fn kill_session(&self, session_id: &str) -> Result<()>;
 }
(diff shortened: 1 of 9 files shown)


BunsDev (changes requested):
The stale-lock recovery can violate maintenance exclusion (`crates/coven-cli/src/maintenance_gate.rs:571-595`).

A lock older than 30 seconds is deleted without proving its owner has exited. After sleep or a stalled operation, another process can replace the lock while the original holder remains in its critical section; the original holder can then unconditionally delete the replacement lock on drop. This permits destructive maintenance to overlap an active session or claim mutation.

Please use an OS-backed lock or ownership tokens/heartbeats so takeover requires a provably dead owner and only the current owner can remove the lock. Add a concurrency test covering stale-lock takeover.

PR #604, opened 2026-08-04
feat(runtime): define versioned harness contract

### LINKED ISSUE #595: runtime: define a versioned harness contract and conformance suite
## Goal
Decouple harness integrations from embedded runner internals with an explicit negotiated contract.

## Scope
- Version capabilities, input, streamed output, lifecycle, cancellation, and errors.
- Define required versus optional extension behavior.
- Publish compatibility and migration policy.
- Add golden fixtures for built-in and external adapters.

## Acceptance
- Unsupported versions fail clearly.
- Unknown optional fields are forward-compatible.
- Required fields fail closed.
- Codex, Claude, Copilot, Coven Code, and one external adapter pass conformance fixtures.
- Cancellation and partial failure semantics are explicit.
- Rollout avoids a flag-day release.

**Bead:** `cave-3qas7.5`  
**Familiar owner:** Astra  
**Program priority:** P1

CompleteDotTech:
Closes #595

## Summary

Harness adapters previously relied on versionless, embedded runner/PTY behavior. That made compatibility, cancellation, partial output, and extension evolution implicit.

## Solution

Define `coven.harness.v1` as a typed, negotiated boundary with fail-closed required capabilities and forward-compatible optional additions. Add synthetic golden conformance vectors for Codex, Claude Code, Copilot CLI, Coven Code, and one external adapter.

## Important changes

- Adds negotiation, typed input/output frames, terminal errors, sequence validation, and cancellation conformance state.
- Makes unsupported versions and unknown required extensions explicit protocol failures.
- Documents lifecycle, migration/no-flag-day policy, fixture privacy, and the canonical adapter contract.

## Validation

- Focused harness-contract suite: 5 passed
- `node --test scripts/cli-docs-test.mjs` (6 passed)
- Exact-head CI: Rust checks (ubuntu-latest), secret guard, contract, package, and smoke checks passed. Windows Rust/npm checks remain in progress.

## Risks / follow-up

This establishes the contract and conformance boundary without forcing a flag-day conversion of existing PTY recipes. Individual live adapters should opt into v1 incrementally with real smoke tests for their advertised lifecycle and cancellation paths.

### DIFF
--- a/crates/coven-cli/src/harness_contract.rs
+++ b/crates/coven-cli/src/harness_contract.rs
@@ -1,0 +1,543 @@
+//! Versioned contract shared by the Coven runner and harness adapters.
+//!
+//! This module deliberately contains no process-spawning code.  The daemon
+//! remains the authority for cwd validation, process ownership, and policy;
+//! adapters negotiate only the capability and event vocabulary they can
+//! faithfully implement.  Keeping this boundary pure makes the wire rules
+//! fixture-testable for bundled and external adapters alike.
+
+use std::collections::BTreeSet;
+use std::fmt;
+
+use serde::{Deserialize, Serialize};
+
+pub const CONTRACT_NAME: &str = "coven.harness";
+pub const CONTRACT_V1: &str = "coven.harness.v1";
+
+/// Extensions understood by the v1 runner.  Base lifecycle, text input,
+/// streamed output, cancellation, and errors are mandatory v1 semantics; the
+/// extension list is for additive behavior only.
+const KNOWN_EXTENSIONS: &[&str] = &["input.image.v1", "output.tool-use.v1"];
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ContractOffer {
+    pub contract: String,
+    pub versions: Vec<String>,
+    #[serde(default)]
+    pub required_extensions: Vec<String>,
+    #[serde(default)]
+    pub optional_extensions: Vec<String>,
+}
+
+impl ContractOffer {
+    pub fn v1() -> Self {
+        Self {
+            contract: CONTRACT_NAME.to_string(),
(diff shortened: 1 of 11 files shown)


BunsDev:
Reviewed as part of a branch/PR-reduction sweep. The contract module itself reads well — negotiation fails closed on version and required-extension mismatch, the conformance state machine enforces ready-before-output, monotonic sequences, and exactly one terminal frame, and the golden fixtures cover all four bundled adapters plus an external one.

**I pushed one fix** (`ab7d197`): the two `[Adapter contract](/ADAPTER-SPEC)` links were dead. `/UPPERCASE` routes resolve to top-level `docs/UPPERCASE.md` (that's how `/HARNESS-ADAPTERS` works), but this PR's page is at `docs/reference/adapter-spec.md`, whose route is `/reference/adapter-spec`. There is no `docs/ADAPTER-SPEC.md`. I also normalized the relative `reference/adapter-spec.md` link in `docs/HARNESS-ADAPTERS.md` to the same absolute form every other cross-reference in `docs/` uses.

Two non-blocking notes for later:

**Cancellation racing a natural completion.** `SessionConformance::accept` rejects a terminal frame whose outcome is not `Cancelled` once `cancellation_requested` is set:

```rust
if self.cancellation_requested.is_some() && *outcome != TerminalOutcome::Cancelled {
    return Err(contract_error("cancellation_outcome_mismatch", ...));
}
```

If a cancel is issued at the moment the adapter finishes its turn, `completed` is the honest outcome, but the contract classifies it as adapter misbehavior. Accepting `completed` after a cancel request — while still recording that cancellation was requested — would model the race without weakening the guarantee that a *forced* stop is never reported as clean completion.

**Version selection is lexicographic.** `negotiate` does `.filter(|v| v.as_str() == CONTRACT_V1).max()`. The filter makes `.max()` a no-op today, so this is correct as written — but if the filter is relaxed when v2 lands, string `max` puts `v9` above `v10`. Worth a comment or a parsed ordering before the second version exists.

Issue #596, opened 2026-08-04
perf: add a bounded dedicated event writer


BunsDev:
## Goal
Replace per-event SQLite open/insert churn with a bounded daemon-owned writer path.

## Scope
- Keep one long-lived writer connection.
- Use a bounded, byte-accounted queue.
- Coalesce adjacent raw output briefly per session.
- Preserve lifecycle, tool, error, and exit events losslessly.
- Flush accepted output before recording session exit.
- Surface pressure and writer failure explicitly.

## Acceptance
- Per-session event ordering is unchanged.
- Exit cannot overtake accepted output.
- Eight noisy sessions remain responsive.
- Connection opens and transactions fall by at least 80%, or an evidence-backed safer target is documented.
- Failure becomes a visible degraded/error state, never silent loss.

**Bead:** `cave-3qas7.2`  
**Familiar owner:** Cody  
**Program priority:** P0  
**Depends on:** `cave-3qas7.1`

PR #606, opened 2026-08-04
feat(daemon): add companion-safe session handoff

### LINKED ISSUE #536: feat(daemon): add companion-safe session handoff and takeover contract
## Problem

Coven Pocket issue OpenCoven/coven-pocket#10 needs to hand a running goal/session between its on-device engine and a paired Coven daemon, in both directions. The current `coven.daemon.v1` surface exposes session list/launch/events/input/kill and travel profile/delta reconciliation, but it has no contract that can safely complete this transfer.

The existing draft `specs/coven-handoff-packet/{PRODUCT,TECH}.md` defines useful portable context, but it is not implemented. Its v1 write path is also Unix-socket-only, so a paired companion cannot emit a handoff. Travel deltas append offline evidence but do not establish session ownership, expose a continuation packet, or reconcile a live workspace.

Without an upstream contract, a client can only copy transcript text into a new prompt. That loses provenance and verification state, cannot detect workspace divergence, and permits both ends to continue writing after an apparent takeover.

## Required contract

- Implement and version a portable handoff packet with task context, current state, touched files, risks, verification, and next action.
- Expose companion-safe read/write handoff routes under the versioned daemon API, with the same structured errors and trust boundary as other paired session traffic.
- Add explicit takeover semantics: one active writer/owner, compare-and-swap generation or lease, source acknowledgement, and idempotent retry behavior.
- Include reconciliation metadata sufficient for a client to detect transcript and workspace divergence before takeover (event cursor/generation, repository identity, commit/branch, dirty/touched-file summary).
- Provide a continuation launch/import operation that consumes a validated packet without treating untrusted packet prose as privileged system instructions.
- Define reverse handoff from daemon to a constrained client that cannot spawn processes and may use a different local checkout path.
- Fail closed on stale generations, changed daemon identity, mismatched repository identity, unresolved dirty-worktree conflicts, or unsupported harness/session modes.

## Acceptance criteria

1. A client can request a handoff snapshot for a live session and receives a stable generation plus redacted packet and workspace metadata.
2. A destination can claim takeover exactly once; stale or duplicate claims are idempotent or rejected with structured errors.
3. The source can no longer accept input after takeover commits.
4. Transcript/event and workspace divergence are reported explicitly; no last-writer-wins overwrite occurs.
5. A destination can launch/import the continuation with provenance linking it to the source session and handoff generation.
6. End-to-end tests cover device-to-daemon, daemon-to-device, concurrent takeover, retry after timeout, daemon restart, and dirty-worktree conflict.

## Related

- OpenCoven/coven-pocket#10
- `specs/coven-handoff-packet/PRODUCT.md`
- `specs/coven-handoff-packet/TECH.md`
- `specs/coven-multi-host-daemon/TECH.md`

CompleteDotTech:
## Summary

Closes #536

Adds a durable `coven.handoff.v1` takeover contract for a companion client. The prior daemon exposed sessions and cursored events, but had no ownership transition: a source and destination could both continue sending input.

## Root cause

Session state and transport cursors existed independently, without a durable compare-and-swap claim, source input fence, workspace identity check, or idempotent continuation record.

## Solution

- Persist offered → claimed → acknowledged → continued handoff state and provenance in SQLite.
- Validate/redact packets, record generation and event cursor, and compare portable Git snapshots before claim.
- Fence source input atomically, require source acknowledgement, and make retry claims/imports idempotent.
- Return packets only in a fixed untrusted-context prelude; document that a remote companion still needs a separately paired authenticated transport.

## Validation

- `cargo fmt --all --check` — passed
- focused compiled handoff test binary — 11 passed
- `cargo clippy -p coven-cli --all-targets --locked -- -D warnings` — passed
- `python3 scripts/check-secrets.py` — passed
- `python3 scripts/check-coven-privacy.py --staged` — passed
- `git diff --check` — passed

## Risks / follow-up

The daemon contract remains local Unix-socket-only. A Pocket/remote transport must be mutually authenticated and paired before it can call these routes; this PR intentionally does not expose them through a remote listener.

### DIFF
--- a/crates/coven-cli/src/api.rs
+++ b/crates/coven-cli/src/api.rs
@@ -20,6 +20,7 @@ use crate::{
     control_plane,
     daemon::DaemonStatus,
     encrypted_artifacts::SensitiveArtifactStore,
+    handoff::{HandoffPacketV1, WorkspaceSnapshot},
     harness::{ConversationHint, HarnessLaunchMode},
     privacy, session_launch, store, ward,
 };
@@ -125,6 +126,7 @@ pub struct HealthCapabilities {
     pub executor_dispatch: bool,
     pub event_cursor: String,
     pub structured_errors: bool,
+    pub session_handoff: bool,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
@@ -247,6 +249,7 @@ pub fn health_response(daemon: Option<DaemonStatus>) -> HealthResponse {
             executor_dispatch: true,
             event_cursor: "sequence".to_string(),
             structured_errors: true,
+            session_handoff: true,
         },
         daemon,
         hub: None,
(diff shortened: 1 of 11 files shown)


BunsDev:
Reviewed as part of a branch/PR-reduction sweep. Two findings, both about how the event cursor is read and what it implies.

### 1. The cursor read is an unbounded full-history scan

`emit_handoff`, `claim_session_handoff`, and `acknowledge_session_handoff` each do:

```rust
let event_cursor = store::list_events(&conn, session_id)?
    .last()
    .map(|event| event.seq)
    .unwrap_or(0);
```

`store::list_events` delegates to `list_events_with_options` with `EventsQueryOptions::default()` — no limit. So every one of these calls materializes **every event row for the session**, including full raw PTY output payloads, into a `Vec<EventRecord>`, in order to read a single integer off the end.

For a session that has been streaming for a while this is megabytes of allocation per handoff request, and it grows without bound over the session's life. `SELECT MAX(seq) FROM events WHERE session_id = ?1` answers this directly, and `list_events_with_options` already exists if a record is genuinely needed.

This is worth fixing regardless, but it gets substantially worse alongside #596/#607, which is specifically about raising retained event volume.

### 2. Cursor equality makes a live-session handoff unclaimable

`emit_handoff` requires the session to be live:

```rust
if session.status != "running" {
    return session_not_live_response(session_id);
}
```

and both claim and ack reject on any cursor movement:

```rust
if current_cursor != handoff.event_cursor {
    return api_error(409, "transcript_diverged", ...);
}
```

Taken together, a handoff succeeds only if the session emits **zero** events between emit and claim. But the session is required to be `running`, and a running harness session is exactly the thing that produces output. Any output in that window permanently 409s the claim — the handoff record can never be claimed, only superseded by a fresh one, which has the same race.

If the intent is "the transcript I snapshotted is still a prefix of the current one," then `current_cursor < handoff.event_cursor` is the divergence condition (the transcript was truncated or rewritten), and forward movement should be allowed with the claimant catching up from `handoff.event_cursor`. If strict equality is genuinely intended, then handoff only works on quiesced sessions and `emit_handoff` should probably say so rather than requiring `running`.

Happy to be wrong on the intent here — but as written I don't see the path where a busy session's handoff gets claimed.

PR #607, opened 2026-08-04
perf: add bounded daemon event writer

### LINKED ISSUE #596: perf: add a bounded dedicated event writer

CompleteDotTech:
## Summary

Closes #596

Live PTY callbacks previously opened a SQLite store and committed one event per callback. Under noisy concurrent sessions this churned connections and transactions on the output-drain path.

## Solution

Add a daemon-owned, long-lived SQLite event writer with a byte-accounted bounded queue. Raw output is non-blocking and may be explicitly rejected under pressure; lifecycle and exit events reserve capacity, commit synchronously, and preserve accepted-output-before-exit ordering.

## Important changes

- Coalesce adjacent same-session raw output in short transaction batches.
- Keep one writer connection and expose transaction/connection/drop counters through the additive health eventWriter response.
- Latch writer failures and expose them as degraded persistence health.
- Drain queued critical acknowledgements when a commit fails, so no producer remains blocked after the worker exits.
- Flush a live output batch after the coalescing window and test the observer's output-before-exit barrier.
- Add ordering, pressure, and eight-session batching tests; the workload demonstrates one connection and at most 16 transactions versus 136 per-callback operations (>80% reduction).
- Document the health contract.

## Validation

- cargo fmt --check
- cargo clippy --workspace --all-targets -- -D warnings
- cargo test --workspace --locked
- node --test scripts/cli-docs-test.mjs
- node --test scripts/onboarding-docs-test.mjs
- python3 scripts/check-secrets.py
- python3 scripts/check-coven-privacy.py --range origin/main...HEAD
- python3 scripts/check-api-contract-docs-test.py
- python3 scripts/check-api-contract-docs.py
- git diff --check

## Risks and follow-up

The queue intentionally drops only raw output under sustained pressure; counts and byte totals remain visible for the daemon lifetime. Native GUI/platform coverage remains CI-owned; validation here ran in WSL.

### DIFF
--- a/crates/coven-cli/src/api.rs
+++ b/crates/coven-cli/src/api.rs
@@ -148,6 +148,10 @@ pub struct HealthResponse {
     /// response is built without store access (e.g. CLI status printing).
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub hub: Option<HubHealth>,
+    /// Daemon-owned event persistence health.  Omitted for status rendering
+    /// paths that do not have a live runtime.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub event_writer: Option<crate::event_writer::EventWriterHealth>,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
@@ -215,6 +219,10 @@ pub trait SessionRuntime {
     fn launch_session(&self, launch: &SessionLaunch) -> Result<()>;
     fn send_input(&self, session_id: &str, payload: &Value) -> Result<()>;
     fn kill_session(&self, session_id: &str) -> Result<()>;
+
+    fn event_writer_health(&self) -> Option<crate::event_writer::EventWriterHealth> {
+        None
+    }
 }
 
 pub struct NoopSessionRuntime;
@@ -250,11 +258,17 @@ pub fn health_response(daemon: Option<DaemonStatus>) -> HealthResponse {
         },
         daemon,
         hub: None,
+        event_writer: None,
     }
 }
 
-fn health_response_with_hub(coven_home: &Path, daemon: Option<DaemonStatus>) -> HealthResponse {
+fn health_response_with_hub(
+    coven_home: &Path,
+    daemon: Option<DaemonStatus>,
+    event_writer: Option<crate::event_writer::EventWriterHealth>,
+) -> HealthResponse {
     let mut response = health_response(daemon);
+    response.event_writer = event_writer;
     if let Ok(summary) = crate::hub::hub_health_summary(coven_home) {
         response.hub = serde_json::from_value(summary).ok();
     }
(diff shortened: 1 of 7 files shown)


BunsDev:
Reviewed as part of a branch/PR-reduction sweep. The core design holds up: one connection on one owner thread, FIFO queue, terminal events reserve capacity and block on a commit acknowledgement so an accepted output chunk can't be overtaken by the exit that follows it. The `event_closed` barrier in `output_observer_with_cleanup` correctly closes the late-stderr window. Lock ordering is queue→health everywhere I checked, so no ABBA.

One gap and three smaller notes.

### Dropped output is invisible in the event stream

Under pressure `record_output` returns `Ok(false)` and the chunk is gone. The only records of that are a stderr line and two **daemon-global** counters (`droppedOutputEvents`, `droppedOutputBytes`) on `/health`.

A client reading `/api/v1/sessions/{id}/events` therefore sees a transcript that looks complete — ordered output events, then a clean `exit` — with a silent hole in the middle. Nothing in the session's own event stream says bytes were lost, and the global counters can't tell you *which* session was truncated or *where*.

Since the queue already reserves `RESERVED_CRITICAL_BYTES` for non-output events, emitting one `output_truncated` event per session per pressure episode (bytes dropped, and coalescing repeats so a sustained episode is one event, not a flood) would come out of that reservation and make the gap self-describing. That seems closer to the PR's own framing that "a rejected raw chunk is never silently forgotten" — right now it isn't forgotten by the *daemon*, but it is invisible to the *consumer*.

### Smaller notes

- **`#[allow(dead_code)]` on `pub fn record` has no justifying comment.** AGENTS.md: "Fix lints; don't `#[allow(...)]` without a justifying comment." One line saying it's the writer-side entry point for lifecycle events pending a caller would satisfy it.
- **Output events now coalesce.** Adjacent same-session chunks in one batch merge into a single event, and the merged event carries the **first** chunk's `created_at`, so its timestamp can trail its last byte by up to the 12ms coalesce window. The 1:1 chunk→event relationship isn't a documented contract, so this is fine, but it's an observable change for anything counting or timestamping output events and may deserve a line in `docs/API-CONTRACT.md` next to the `eventWriter` block.
- **One condvar serves two wait conditions.** `available` is waited on both by producers blocked for capacity in `enqueue_critical` and by the worker blocked for items in `take_batch`, while enqueue signals with `notify_one`. I worked through the lost-wakeup case and believe it's safe — a producer can only be capacity-blocked while `queued_bytes` is high, which means either items are non-empty or a batch is in flight, so the worker is never the waiter that gets skipped — and the 12ms `wait_timeout` bounds the coalesce path anyway. Two condvars (or `notify_all` on enqueue) would remove the need for that argument.

Also worth noting for the release notes: `serve_forever` now propagates `try_with_coven_home`, so a daemon that cannot open the store fails to start rather than starting with event persistence broken. That's the better behavior, and `recover_orphaned_sessions` already opens the store first so the failure point barely moves — just a visible change in symptom.

PR #608, opened 2026-08-04
feat(store): automate bounded retention health

### LINKED ISSUE #597: reliability: automate event retention and expose SQLite health
## Goal
Keep the event ledger convergent and surface storage pressure before it becomes an incident.

## Scope
- Prune expired events in small scheduled transactions.
- Keep full VACUUM operator-triggered.
- Rotate recovery logs.
- Report database/WAL size, checkpoint/prune age, oldest retained event, writer backlog, and free disk.
- Enforce maintenance safety watermarks.

## Acceptance
- A 30-day synthetic workload converges to a stable disk footprint.
- Interrupted pruning preserves event and FTS integrity.
- Maintenance stays inside the launch-latency budget.
- Pressure is visible before exhaustion.
- Automatic startup never runs full VACUUM.

**Bead:** `cave-3qas7.4`  
**Familiar owner:** Salem  
**Program priority:** P1  
**Depends on:** `cave-3qas7.2`

CompleteDotTech:
Closes #597

## Summary

Automates bounded SQLite event and artifact retention and makes storage pressure observable through daemon health.

## Root cause

Retention was manual and the health endpoint did not expose database, WAL, free-space, or maintenance state. This left long-running installations without a bounded maintenance path or an operator signal before storage pressure became disruptive.

## Changes

- Run small, transactional retention batches after daemon startup without automatic VACUUM.
- Preserve event/FTS consistency within each prune transaction and use a passive WAL checkpoint only once the WAL crosses its threshold.
- Honor the same cached Settings, `privacy.toml`, and environment retention precedence as `coven logs prune`.
- Add SQLite storage, maintenance watermark, free-disk, and writer-backlog fields to `GET /api/v1/health`.
- Block maintenance below the 256 MiB free-space safety watermark without opening or writing SQLite.
- Rotate recovery logs at 4 MiB with three archives, retaining valid older history across a partial prior rotation.
- Document the health contract and explicit operator compaction path.

## Validation

- `cargo fmt --check`
- Focused retention, configured-retention, interrupted-transaction/FTS, synthetic steady-state, recovery-log, and no-write-below-watermark tests
- `python3 scripts/check-coven-privacy.py --range origin/main...HEAD`
- `python3 scripts/check-api-contract-docs.py`

## Risks and follow-up

- The current direct-write implementation accurately reports a zero writer backlog. Issue #596 can populate the same stable health fields from its dedicated writer queue.
- Automatic `VACUUM` remains intentionally excluded because it can monopolize SQLite and affect session-launch latency; operators retain the explicit `coven vacuum` command.

### DIFF
--- a/crates/coven-cli/src/api.rs
+++ b/crates/coven-cli/src/api.rs
@@ -289,10 +289,8 @@ fn health_response_with_hub(
     if let Ok(summary) = crate::hub::hub_health_summary(coven_home) {
         response.hub = serde_json::from_value(summary).ok();
     }
-    response.storage = Some(
-        store::storage_health(coven_home)
-            .unwrap_or_else(|error| store::unavailable_storage_health(error)),
-    );
+    response.storage =
+        Some(store::storage_health(coven_home).unwrap_or_else(store::unavailable_storage_health));
     response
 }
 
--- a/crates/coven-cli/src/api.rs
+++ b/crates/coven-cli/src/api.rs
@@ -154,6 +154,11 @@ pub struct HealthResponse {
     /// paths that do not have a live runtime.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub event_writer: Option<crate::event_writer::EventWriterHealth>,
+    /// Local SQLite pressure and bounded-maintenance state. This remains
+    /// present when collection fails so health consumers can distinguish a
+    /// storage problem from a daemon that is simply not running.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub storage: Option<store::StorageHealth>,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
@@ -270,6 +275,7 @@ pub fn health_response(daemon: Option<DaemonStatus>) -> HealthResponse {
         daemon,
         hub: None,
         event_writer: None,
+        storage: None,
     }
 }
 
@@ -283,6 +289,10 @@ fn health_response_with_hub(
     if let Ok(summary) = crate::hub::hub_health_summary(coven_home) {
         response.hub = serde_json::from_value(summary).ok();
     }
+    response.storage = Some(
+        store::storage_health(coven_home)
+            .unwrap_or_else(|error| store::unavailable_storage_health(error)),
+    );
     response
 }
 
@@ -6726,6 +6736,7 @@ mod tests {
         assert!(response.capabilities.structured_errors);
         assert_eq!(response.daemon, None);
         assert_eq!(response.hub, None);
+        assert_eq!(response.storage, None);
     }
 
     #[test]
@@ -6755,6 +6766,9 @@ mod tests {
         assert!(response.body.contains(r#""executorDispatch":true"#));
         assert!(response.body.contains(r#""role":"hub""#));
         assert!(response.body.contains(r#""structuredErrors":true"#));
+        assert!(response.body.contains(r#""storage":{"status""#));
+        assert!(response.body.contains(r#""writerBacklogEvents":0"#));
+        assert!(response.body.contains(r#""freeDiskBytes""#));
         Ok(())
     }
 
--- a/crates/coven-cli/src/daemon.rs
+++ b/crates/coven-cli/src/daemon.rs
@@ -2080,15 +2080,17 @@ fn rotate_recovery_log(path: &Path, incoming_bytes: u64, max_bytes: u64, backups
 
     for index in (1..=backups).rev() {
         let destination = PathBuf::from(format!("{}.{}", path.display(), index));
-        if index == backups {
-            let _ = std::fs::remove_file(&destination);
-        }
         let source = if index == 1 {
             path.to_path_buf()
         } else {
             PathBuf::from(format!("{}.{}", path.display(), index - 1))
         };
         if source.exists() {
+            // Windows does not replace an existing destination for
+            // `std::fs::rename`, unlike Unix. Only remove a slot immediately
+            // before replacing it, so a lone older archive survives a partial
+            // prior rotation where its newer source is absent.
+            let _ = std::fs::remove_file(&destination);
             let _ = std::fs::rename(source, destination);
         }
     }
@@ -5569,6 +5571,21 @@ mod tests {
         assert_eq!(
             std::fs::read_to_string(format!("{}.1", path.display()))?,
             "abcdefgh"
+        );
+        assert_eq!(
+            std::fs::read_to_string(format!("{}.2", path.display()))?,
+            "12345678"
+        );
+
+        // A partial prior rotation can leave an older archive without its
+        // newer neighbor. Keep the surviving history rather than deleting it
+        // merely because there is no source to shift into its slot.
+        std::fs::remove_file(format!("{}.1", path.display()))?;
+        std::fs::write(&path, "ijklmnop")?;
+        rotate_recovery_log(&path, 4, 10, 2);
+        assert_eq!(
+            std::fs::read_to_string(format!("{}.1", path.display()))?,
+            "ijklmnop"
         );
         assert_eq!(
             std::fs::read_to_string(format!("{}.2", path.display()))?,
--- a/crates/coven-cli/src/daemon.rs
+++ b/crates/coven-cli/src/daemon.rs
@@ -2041,15 +2041,56 @@ pub fn daemon_recovery_log_path(coven_home: &Path) -> PathBuf {
     coven_home.join("daemon-recovery.log")
 }
 
+const DAEMON_RECOVERY_LOG_MAX_BYTES: u64 = 4 * 1024 * 1024;
+const DAEMON_RECOVERY_LOG_BACKUPS: u8 = 3;
+
+fn recovery_log_lock() -> &'static Mutex<()> {
+    static LOCK: std::sync::OnceLock<Mutex<()>> = std::sync::OnceLock::new();
+    LOCK.get_or_init(|| Mutex::new(()))
+}
+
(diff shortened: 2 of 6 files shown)


BunsDev (commented):
Superseded — this review was re-submitted as a blocking one, with the same content plus the non-blocking notes, at https://github.com/OpenCoven/coven/pull/608#pullrequestreview-4859637197. GitHub does not allow deleting a submitted review, so this stub stands in its place.

BunsDev (changes requested):
The overall design judgment is right — no automatic `VACUUM`, transactional prune so the FTS delete trigger stays consistent, `PASSIVE` over `TRUNCATE`, the watermark check before touching SQLite, and no `COVEN_HOME` path disclosure across the socket boundary. Retention precedence does match `prune_logs_command` (`main.rs:2988-2993`), `.max(1)` clamp included.

The four items below are what I'd want resolved before merge; the smaller notes after them are non-blocking.

### 1. The artifact prune full-scans the table every 60 seconds

`prune_sensitive_artifacts_bounded`:

```sql
WHERE expires_at < ?1 OR created_at < ?2 ORDER BY created_at, rowid LIMIT ?3
```

This PR adds `idx_events_created_at` because "the session-scoped index above cannot serve that scan" — the identical problem exists on `sensitive_artifacts` and no matching index came with it. The existing indexes are `(session_id, created_at)` and `(expires_at)` (`store.rs:488-492`); there is no standalone `created_at` index, and the `OR` plus `ORDER BY created_at` defeats what indexing there is. Every tick pays a full scan and sort.

Add the matching `created_at` index, or split this into two bounded deletes (one per predicate) so each is index-served.

### 2. Fixed batch, no drain loop, no catch-up

One 500-row batch per 60s is 8.3 rows/s, ~720k rows/day. A pre-existing backlog on first upgrade drains at that fixed rate regardless of size, and sustained ingestion above ~8 events/s means retention never converges — the store grows unboundedly while the daemon reports it is pruning. Looping batches until a pass returns fewer than `limit` (with a per-tick cap), or scaling the batch to the observed backlog, would close this.

Together with (3), this is what undercuts the PR's own bounded-storage guarantee.

### 3. `status` cannot see retention lag

`storage_health` computes `status` from free disk, WAL size, and last error only. A pruner stalled by (2) reports `"ok"` while `oldestRetainedEventAt` drifts arbitrarily far past the cutoff. Given the stated goal — "an operator signal before storage pressure became disruptive" — the most direct lag signal is collected and then never consulted. Comparing `oldest_retained_event_at` against `retention_cutoff` and warning past some margin would make the field earn its place.

### 4. Two tests read real machine state

`scheduled_maintenance_prunes_retention_without_vacuuming` and `thirty_day_synthetic_workload_converges_to_a_stable_store_size` call `run_scheduled_maintenance` directly, which reads real `fs2::available_space` and real `privacy::load_with_settings(..., settings::cached())` — including `COVEN_LOG_RETENTION_DAYS` / `COVEN_RAW_ARTIFACT_RETENTION_DAYS` from the process env (`privacy.rs:45-46`). A developer or CI runner with either var set, or under 256 MiB free, gets a silently different or failing test; the `retained <= 31` assertion depends directly on the default 30. Both should use `run_scheduled_maintenance_with_config_and_free_disk`, as the third test already does.

### Smaller notes (non-blocking)

- **The 16 MiB checkpoint is close to dead code.** Nothing in the crate sets `wal_autocheckpoint`, so SQLite's default 1000-page (~4 MiB) autocheckpoint already runs on commit. The WAL only exceeds 16 MiB when a long-lived reader blocks checkpointing — and `PASSIVE` is blocked by that same reader. `lastCheckpointAt` will be null in practice; the `warning` gate on WAL size is the part actually earning its keep.
- **An idle daemon is never quiescent.** Every tick writes `last_prune_at` and re-clears `last_error` — two `store_meta` writes per minute forever, even with nothing to prune. Writing `last_error` only on transition would avoid half of that.
- **`lastMaintenanceError` is a constant.** `record_maintenance_error(&home, "maintenance pass failed")` discards `{error:#}`, so the field is effectively a boolean; the real error reaches only the recovery log.
- **`maintenanceBlocked: true` on read failure** conflates "below watermark" — what `docs/reference/api.md` documents it as — with "could not determine."
- **Watermark bypass on `fs2` failure.** If `available_space` errors, `run_scheduled_maintenance` returns `Err`, and the daemon handler then calls `record_maintenance_error`, which opens the store and writes — the one path meant to avoid writes under disk pressure.
- **Startup cost on upgrade.** `idx_events_created_at` is built inside `initialize_store`'s `BEGIN IMMEDIATE` (`store.rs:412`), so an existing install with a large `events` table pays a one-time blocking index build at daemon start. Worth a release note.
- **Test value.** `bounded_event_pruning_keeps_fts_consistent_across_an_interruption` hand-rolls its own `DELETE` and drops the transaction, so it demonstrates SQLite rollback rather than `prune_events_older_than_bounded`'s atomicity. The second half — the real call plus `pragma_integrity_check` — is the part carrying the weight.
- **Visibility.** `prune_events_older_than_bounded` and `MAINTENANCE_MIN_FREE_DISK_BYTES` are `pub` with no consumer outside `store.rs` and its tests.

### Verdict

Nothing here corrupts data or breaks the API contract — the serde additions are backward-compatible and the transactional/FTS handling is right. Items 1-4 above are the gate.

BunsDev (changes requested):
Re-reviewed the current head. The prior four design blockers are addressed: the prune path is indexed/bounded, maintenance can converge, retention lag affects status, and disk metrics are injectable in tests.

Three current blockers remain:

1. The exact head does not compile. CI reports two E0308 errors at `crates/coven-cli/src/store.rs:3191` (`usize` batch counts compared with `i64` constants) and E0282 at `store.rs:3238` (`oldest_retained_event_at` needs an explicit type).
2. This branch conflicts with current `main`, where #607/#596 has already landed the bounded event writer and its real `EventWriterHealth`. Rebase and reconcile the health contract: the new `writerBacklogEvents` / `writerBacklogBytes` values cannot remain hard-coded zero, and the docs saying #596 will populate them are now stale.
3. Commits `fbc2156`, `eee01c0`, `6b1d63b`, and `22bc2e6` are missing `Signed-off-by:` trailers required by DCO.

Please rebase, integrate the real writer-health values, sign off every commit, and rerun the full repository gates.

BunsDev:
Rebased onto current `main` and fixed a build break. Detail below, because the PR's status was misleading in both directions — it was closer to done than it looked, and also more broken.

## All four blocking review items are addressed

Your six follow-up commits resolved them; I verified each against the code rather than the commit titles:

| item | resolution |
|---|---|
| 1. Artifact prune full-scans every 60s | The `OR` is split into a `UNION` of two separately-servable branches, **and** `idx_sensitive_artifacts_created_at` was added — in both the fresh schema and the migration path, so existing installs get it too. |
| 2. Fixed batch, no drain loop | `MAINTENANCE_MAX_BATCHES_PER_TICK = 10` with a convergence loop that breaks once both batches come back short. Bounded catch-up rather than an unbounded drain. |
| 3. `status` cannot see retention lag | `retention_lagging` compares `oldest_retained_event_at` against `retention_cutoff` and drives `status` to `warning`. |
| 4. Two tests read real machine state | `run_scheduled_maintenance_with_free_disk` now delegates to the config-injecting helper with `&PrivacyConfig::default()`, so the tests no longer pick up `COVEN_LOG_RETENTION_DAYS` from the process environment. |

## The branch had never compiled

This is the part worth flagging. Three errors — `E0308` twice and `E0282` — were reported by **this PR's own CI at 04:40**, before any rebase. They read as fallout from `main` being broken at the time (it was, separately, and #614/#616 fixed that), so they went unexamined.

Both are in the code added for the review items:

- **`E0308` ×2.** `prune_sensitive_artifacts_bounded` and `prune_events_older_than_bounded` report rows deleted as `usize`, while `MAINTENANCE_ARTIFACT_BATCH_SIZE` / `MAINTENANCE_EVENT_BATCH_SIZE` are `i64` because they bind directly to SQL `LIMIT`. The convergence loop compared them directly. Fixed by comparing in `i64`, so the value bound to SQLite and the value compared stay the same type.
- **`E0282`.** `SELECT MIN(created_at)` is NULL on an empty ledger, and `row.get` cannot infer `Option<String>` from the `retention_lagging` chain alone. Annotated explicitly.

Both sites carry a comment explaining the constraint, since either is easy to "simplify" back into a break.

## Rebase notes

`main` had moved 23 commits. Three conflicts, all resolved as both/and rather than either/or:

- **`api.rs`** — `HealthResponse` carries both `event_writer` (#607) and `storage` (this PR).
- **`store.rs`** — the handoff tables (#606) and `idx_events_created_at`. I checked that index is genuinely absent from `main` before keeping it.
- **`docs/daemon/health.md`** — kept `main`'s `<covenHome>` placeholder over `$COVEN_HOME`.

One doc comment corrected while resolving: `StorageHealth` said the dedicated writer was "planned in #596" and would replace `writer_backlog_events` / `writer_backlog_bytes`. That writer landed in #607 while this PR waited. The comment now says those fields report zero and that real queue depth is `eventWriter.queuedBytes`, with the wiring called out as follow-up rather than guessed at mid-rebase.

## Verification

```
cargo fmt --check                                        clean
cargo clippy --workspace --all-targets -- -D warnings     clean
cargo test --workspace --locked                           1745 + 92 passing, 0 failed
python3 scripts/check-secrets.py                          no current-tree or history findings
```

All seven of your commits keep their authorship; only the compile fix is attributed separately.

BunsDev:
Superseded by #621, which ports this work onto current main, integrates exact live event-writer health, preserves contributor attribution, and resolves the DCO, compile, and merge-conflict blockers.

Issue #615, opened 2026-08-05
benchmark-chaos fails on Linux CI: no output event, gating the perf job


BunsDev:
`CLI performance baseline` has failed on every `main` run since d205f76. The step is:

```
node scripts/benchmark-chaos.mjs --binary target/debug/coven --output coven-chaos.json
benchmark-chaos: harness output event did not arrive: no output event yet
##[error]Process completed with exit code 1.
```

## What has been ruled out

- **Not the #603/#607 compile break.** That was `event_writer_health` landing in the wrong impl block, fixed in #614. Both `Rust checks` jobs pass now; this one still fails.
- **Not a #607 latency regression.** Measured pre-#607 (622ed28) against post-#607, three alternating runs each on an idle machine:

  | build | sessions_1 p50 | sessions_8 p50 | sessions_32 p50 |
  |---|---|---|---|
  | pre-#607 | 834 / 803 / 642 ms | 863 / 633 / 818 ms | 1738 / 1281 / 1294 ms |
  | post-#607 | 735 / 1181 / 607 ms | 708 / 1271 / 1267 ms | 1817 / 1401 / 1673 ms |

  Run-to-run variance exceeds the difference between builds.
- **Not caused by the commit where it first appears.** d205f76 (#610) touched only `.github/actions/branch-triage/*` — three files, none of which CI's perf job reads. It also passed at 114f9c8, which already contained #607.
- **Not concurrency overload.** From the CI timestamps the step dies ~8s in, which is daemon start plus one exhausted `waitForOutputEvent`. That is the **`sessions_1`** scenario — a single session.
- **Not reproducible on macOS.** Locally the full 1/8/32 matrix passes and writes a valid report.

`benchmark-cli.mjs` — same helpers, same `waitForOutputEvent`, same `harness: 'codex'` fixture pattern — passes in the same job, immediately before, taking 52s.

## Hypothesis (unconfirmed)

The two scripts build their fixture harness differently. `benchmark-cli.mjs`:

```js
await writeFile(executable, '#!/bin/sh\nprintf "benchmark output\\n"\n', { mode: 0o700 });
await chmod(executable, 0o700);
```

`benchmark-chaos.mjs`:

```js
await writeFile(harness, '#!/bin/sh\nprintf "COVEN_BENCHMARK_READY\\n"\n...', { mode: 0o700 });
```

[...]

PR #616, opened 2026-08-05
ci: stop the chaos baseline from gating merges, and make its failure diagnosable

BunsDev:
Unblocks `main`, which has been red since d205f76 on this one job. Refs #615.

## Two changes

**1. The chaos collection step no longer gates merges.**

`README.md` already described these baselines as non-gating:

True of thresholds, but not of errors — any failure inside the script failed the step and therefore the job. `continue-on-error: true` makes the collection step match the documented intent.

What still gates is the deterministic part: `benchmark-chaos.test.mjs` runs in `Validate benchmark runner` and is untouched. Only wall-clock collection becomes advisory. The flag is narrow and carries a comment pointing at #615, so it gets removed when the real failure is fixed rather than quietly becoming permanent.

**2. The failure is now self-describing.**

- Scenario, session status, and the event kinds that actually arrived are reported on timeout.
- The fixture harness is asserted owner-executable after writing, with an explicit `chmod`. `writeFile`'s `mode` applies only on creation and is masked by umask (verified: under umask `0o177`, `mode: 0o700` yields `600`), so a non-executable harness would spawn nothing and look identical to a timeout. Not the cause here — the assertion never fires — but no longer silent.
- A larger launch-wait budget, retained as a hang bound. It does not affect the reported metric, which is measured from the clock rather than the poll count.

## What the diagnostics established

The first CI run on this branch passed; a second run on an empty commit, same tree, failed:

```
benchmark-chaos: sessions_32: harness output event did not arrive: no output event yet
                 — status=running events=[none]
```

Intermittent, and it invalidated three hypotheses — all of them mine:

| Hypothesis | Verdict |
|---|---|
| #607 latency regression | **No.** Pre/post-#607 across alternating runs; between-run variance exceeds the difference. |
| Missing `chmod` on the fixture harness | **No.** The assertion added here never fires. |
| Cold-start budget on `sessions_1` | **No.** CI artifact: `sessions_1` 65 ms, `sessions_8` 78 ms, `sessions_32` 222 ms. Raising 4 s → 60 s did not help. |

The failing scenario is **`sessions_32`**, not `sessions_1`. Failures landing ~8 s into the step were three daemon startups plus a 4 s timeout on the third scenario — I had assumed a slow first scenario without ever measuring CI, where it is ~100× faster than my machine.

## The underlying failure, unfixed

`status=running` with `events=[none]`: under 32-way concurrency a live session has recorded **zero events of any kind** after 60 s. Nothing persisted at all. Two candidates, neither confirmed:

1. A daemon concurrency problem on the event path — #607 moved persistence to a queued worker thread, and 32 concurrent PTY drains is the first real contention it sees.
2. 32 simultaneous PTY spawns exceed a two-core runner, making the scenario unrunnable in CI as designed.

Deliberately not fixed here. This PR makes `main` green and makes the next investigation cheap; #615 tracks the cause.

## Verification

```
node --test scripts/benchmark-cli.test.mjs scripts/benchmark-chaos.test.mjs
ℹ pass 52   ℹ fail 0
```

Workflow YAML parsed and asserted: `job=performance-baseline step=Collect concurrent runtime baseline artifact continue-on-error=True`.

One thing to watch on this run: the upload step keeps `if-no-files-found: error`, and `coven-chaos.json` will be absent whenever collection fails. That should be fine since the other two artifacts exist and the check is on the aggregate match, but this run is the confirmation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,12 @@ jobs:
         run: node --test scripts/benchmark-cli.test.mjs scripts/benchmark-chaos.test.mjs
       - name: Collect non-gating baseline artifact
         run: node scripts/benchmark-cli.mjs --binary target/debug/coven --iterations 3 --output coven-perf.json
+      # Trend collection only, and intermittently unable to run 32 concurrent
+      # sessions on a two-core runner (#615) — a live session there can record
+      # no events at all. Collection must not gate merges; the deterministic
+      # unit tests above still do. Drop `continue-on-error` once #615 is fixed.
       - name: Collect concurrent runtime baseline artifact
+        continue-on-error: true
         run: node scripts/benchmark-chaos.mjs --binary target/debug/coven --output coven-chaos.json
       - name: Collect deterministic TUI scheduling metric
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,7 @@ jobs:
       # unit tests above still do. Drop `continue-on-error` once #615 is fixed.
       - name: Collect concurrent runtime baseline artifact
         continue-on-error: true
-        run: node scripts/benchmark-chaos.mjs --binary target/debug/coven --output coven-chaos.json
+        run: node scripts/benchmark-chaos.mjs --binary /nonexistent/coven --output coven-chaos.json
       - name: Collect deterministic TUI scheduling metric
         run: |
           set -o pipefail
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,7 @@ jobs:
       # unit tests above still do. Drop `continue-on-error` once #615 is fixed.
       - name: Collect concurrent runtime baseline artifact
         continue-on-error: true
-        run: node scripts/benchmark-chaos.mjs --binary /nonexistent/coven --output coven-chaos.json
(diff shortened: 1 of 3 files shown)


BunsDev:
## Non-gating behaviour verified, not assumed

The green run before this one passed because the flaky step happened to succeed, which proved nothing about the failure path. I pushed a temporary probe pointing the step at a nonexistent binary to force a deterministic failure, then reverted it.

With the chaos step genuinely failing:

```
benchmark-chaos: spawnSync /nonexistent/coven ENOENT
```

the job still reported **success**, and the upload step — which keeps `if-no-files-found: error` — completed with `coven-chaos.json` absent:

```
Uploading artifact: coven-cli-performance-baseline.zip
Artifact coven-cli-performance-baseline has been successfully uploaded! Final size is 1095 bytes.
```

So both risks flagged on this PR are now measured:

| Risk | Result |
|---|---|
| `continue-on-error` keeps the job green when collection fails | Confirmed |
| `if-no-files-found: error` tolerates a missing `coven-chaos.json` | Confirmed — the aggregate match is what counts |

Worth noting for anyone reading the API later: GitHub reports a `continue-on-error` step's `conclusion` as `success` even when it failed. The step-level conclusion field is not a reliable signal here; the job log is.

The probe is reverted. Net diff against `main` is `ci.yml` +5, `README.md` +7, `scripts/benchmark-chaos.mjs` +54/-2.

PR #618, opened 2026-08-05
fix: prevent maintenance gate lock takeover

### LINKED ISSUE #612: Maintenance gate: stale-lock takeover has no owner identity, and Drop deletes by path
Follow-up from #603, which merged at 622ed28 with this review finding unresolved. Filing so it isn't lost.

`GateLock` in `crates/coven-cli/src/maintenance_gate.rs` guards mutations of the gate's own metadata (owner lease, writer intents) via `with_lock`. Four properties combine badly.

## 1. The lock file carries no owner identity

`GateLock::acquire` creates the file and immediately discards the handle without writing anything:

```rust
match fs::OpenOptions::new().write(true).create_new(true).open(&path) {
    Ok(_) => return Ok(Self { path }),
```

The file is empty. Nothing records which process, thread, or generation holds it.

## 2. Staleness measures acquisition time, not liveness

`is_stale` reads the file's mtime, and nothing ever refreshes it — no heartbeat, no touch, no rewrite. So `LOCK_STALE_AFTER` (30s) is measured from the moment the lock was taken, and a holder that is *alive but stalled* is indistinguishable from one that died.

## 3. Takeover deletes without proving the owner exited

```rust
if is_stale(&path, LOCK_STALE_AFTER) {
    let _ = fs::remove_file(&path);
}
```

No PID liveness check, no `kill(pid, 0)`, no lease generation. Given (1) there is nothing to check *against*.

## 4. `Drop` deletes by path, unconditionally

```rust
impl Drop for GateLock {
    fn drop(&mut self) {
        let _ = fs::remove_file(&self.path);
    }
}
```

It removes whatever currently sits at that path — not necessarily the file this `GateLock` created.

## The failure

1. Process A acquires the lock and enters its critical section.
2. A stalls past 30s — laptop sleep, a hung network filesystem, SIGSTOP, heavy IO contention, a debugger. The critical sections are short in the normal case, so this needs a stall; it does not need a crash.
3. Process B finds the lock stale, deletes A's file, creates its own, and enters the critical section. **A and B now mutate gate state concurrently** — which is precisely the exclusion the gate exists to provide.
4. A resumes and finishes. Its `Drop` deletes **B's** lock file.
5. Process C acquires immediately while B is still inside. The corruption is now self-sustaining: every subsequent holder can be evicted by the previous one's `Drop`.

Step 4 is the part that makes this more than a rare race — once a single takeover occurs, mutual exclusion stays broken for every holder that follows, without any further stall.

## The fix pattern already exists in this file

The owner and writer leases in the same module do this correctly — they carry a generation token, renew on `RENEW_EVERY` (20s) against `WRITER_TTL` (90s) / `OWNER_TTL` (120s), and `release_writer` verifies ownership before removing:

```rust
if intent.generation == generation {
    let _ = fs::remove_file(path);
}
```

`GateLock` should do the same, or move to an OS-backed advisory lock (`flock`/`LockFileEx`) where the kernel releases on process death and takeover cannot be forged. At minimum:

- write a generation token (and pid) into the lock file on acquire;
- have `Drop` re-read and only unlink when the token still matches;
- either heartbeat the mtime, or treat a live pid as not-stale regardless of age.

## Test worth adding

A concurrency test covering takeover directly: hold a lock, force its mtime past `LOCK_STALE_AFTER` without exiting, let a second acquirer take over, then drop the first and assert the second's lock survives and the second still holds exclusion.

Context: [review on #603](https://github.com/OpenCoven/coven/pull/603).

BunsDev:
Summary: replaces stale-file takeover with existing cross-platform fs2 advisory locking; uses validated no-follow/single-link lock opening; preserves five-second contention timeout and maintenance API behavior; adds live contention/release/stale-mtime/symlink/hard-link regression tests.

Tests:
- cargo fmt --check
- cargo clippy --workspace --all-targets -- -D warnings
- cargo test --workspace --locked
- python scripts/check-secrets.py

Upgrade note: restart long-running Coven daemons after upgrading so all processes use the advisory-lock protocol.

Closes #612

### DIFF
--- a/crates/coven-cli/src/maintenance_gate.rs
+++ b/crates/coven-cli/src/maintenance_gate.rs
@@ -12,10 +12,11 @@ use std::{
     path::{Path, PathBuf},
     sync::{Arc, Condvar, Mutex},
     thread,
-    time::{Duration, SystemTime, UNIX_EPOCH},
+    time::{Duration, Instant, SystemTime, UNIX_EPOCH},
 };
 
 use anyhow::{Context, Result};
+use fs2::FileExt;
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
@@ -23,7 +24,6 @@ const GATE_DIR: &str = "coven-maintenance-gate";
 const OWNER_FILE: &str = "owner";
 const WRITERS_DIR: &str = "writers";
 const LOCK_FILE: &str = "lock";
-const LOCK_STALE_AFTER: Duration = Duration::from_secs(30);
 const LOCK_WAIT: Duration = Duration::from_secs(5);
 const WRITER_TTL: Duration = Duration::from_secs(90);
 const OWNER_TTL: Duration = Duration::from_secs(120);
@@ -553,36 +553,37 @@ impl OwnerLease {
     }
 }
 
+#[derive(Debug)]
 struct GateLock {
-    path: PathBuf,
+    file: fs::File,
 }
 
 impl GateLock {
     fn acquire(path: PathBuf) -> Result<Self> {
-        let started = SystemTime::now();
+        Self::acquire_with_wait(path, LOCK_WAIT)
+    }
+
+    fn acquire_with_wait(path: PathBuf, wait: Duration) -> Result<Self> {
+        let file = fs::OpenOptions::new()
+            .read(true)
+            .write(true)
+            .create(true)
+            .truncate(false)
+            .open(&path)
+            .with_context(|| format!("failed to open {}", path.display()))?;
+        let started = Instant::now();
         loop {
-            match fs::OpenOptions::new()
-                .write(true)
-                .create_new(true)
-                .open(&path)
-            {
-                Ok(_) => return Ok(Self { path }),
-                Err(error) if error.kind() == std::io::ErrorKind::AlreadyExists => {
-                    if is_stale(&path, LOCK_STALE_AFTER) {
-                        let _ = fs::remove_file(&path);
-                    }
-                    if SystemTime::now()
-                        .duration_since(started)
-                        .unwrap_or_default()
-                        > LOCK_WAIT
-                    {
+            match file.try_lock_exclusive() {
+                Ok(()) => return Ok(Self { file }),
+                Err(error) if crate::state_lock::is_lock_contended(&error) => {
+                    if started.elapsed() >= wait {
                         return Err(GateError::Contended.into());
                     }
                     thread::sleep(Duration::from_millis(10));
                 }
                 Err(error) => {
                     return Err(error)
-                        .with_context(|| format!("failed to acquire {}", path.display()))
+                        .with_context(|| format!("failed to acquire {}", path.display()));
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


BunsDev:
Reviewed as the author of #612. This fixes all four defects structurally rather than patching around them.

| #612 defect | resolution |
|---|---|
| Lock file carries no owner identity | Advisory lock — the kernel owns it, so identity is implicit in the open file description. Nothing to forge. |
| Staleness measured acquisition time, and nothing refreshed the mtime | `is_stale` and `LOCK_STALE_AFTER` deleted outright. No staleness concept remains. |
| Takeover unlinked without proving the owner had exited | No takeover path exists. `try_lock_exclusive` fails while a holder is live; the kernel releases on process death. |
| `Drop` unlinked by path, so a departing holder deleted its *successor's* lock | `Drop` calls `FileExt::unlock(&self.file)` — it can only release the handle it holds. |

That last row is what made the original bug self-sustaining: once one takeover happened, exclusion stayed broken for every subsequent holder with no further stall. It is now impossible by construction.

## Verified beyond the description

- **Nothing else depends on the lock file's absence.** It is never unlinked now; the remaining `remove_file` calls are all owner/writer *lease* records, which is correct — those are leases, not the mutex. `eval_loop.rs`'s `RUN_LOCK_FILE` is an unrelated lock.
- **`open_lock_file` is genuinely hardened**, not just reused: `cap_std` with `FollowSymlinks::No`, create-without-truncate, link validation, mode `0600`. That is what the symlink and hard-link tests pin, and the symlink test correctly asserts the outside target is left untouched.
- **`is_lock_contended` separates contention from real errors** (`WouldBlock` or fs2's errno), so a genuine I/O failure is not retried for five seconds and then mislabelled `Contended`.
- **`SystemTime` → `Instant`** is an unadvertised improvement worth noting. The old loop used `duration_since(...).unwrap_or_default()`, which yields `0` under backwards clock skew and could spin well past `LOCK_WAIT`. `Instant` is monotonic.
- **The regression test I asked for is present.** `stale_mtime_does_not_allow_live_gate_lock_takeover` backdates the mtime by 120s — 4× the old 30s threshold — while the lock is held, and asserts contention rather than takeover. That is exactly the case the old code got wrong.

The `acquire_with_wait(.., Duration::ZERO)` seam is a nice touch: it makes contention deterministic in tests without a sleep, while `acquire` keeps the production five-second window.

## One observation, non-blocking

Every new test is same-process. That is still valid coverage — `flock` is per-open-file-description, so two `open_lock_file` calls in one process genuinely contend — but nothing exercises the cross-process scenario from the original report. Since `Drop` can no longer touch another holder's lock, the cascade is structurally unreachable, so I do not think this is worth blocking on. Worth knowing if anyone later reworks `GateLock` and leans on these tests as full coverage.

The upgrade note is correct and worth carrying into the release notes: mixed old/new processes would not interlock, since one unlinks paths while the other holds advisory locks.

Looks right to me. All 12 checks green.

(Posted as a comment rather than an approving review — GitHub will not let this account approve its own pull request.)

PR #629, opened 2026-08-05
docs: plan Psyche W2 G2 foundation

### LINKED ISSUE #627: docs: plan Psyche W2 G2 foundation
## Summary
Create the bounded follow-on W2 implementation plan that closes G2 after the Psyche bootstrap merges.

## Scope
- canonical surface-neutral schema types
- SQLite migrations, transactions, retention, and quarantine
- fake Coven and fake surface boundaries
- state-machine, property, migration, and crash/restart evidence

No W3 identity resolution, W4 orchestration policy, real Coven integration, Telegram adapter, or capability activation.

BunsDev:
## Summary
- gate implementation on Psyche bootstrap PR #1 merging and `main` restoration
- define typed canonical records, strict version denial, SQLite migration/store work, and behavior-level Coven/surface fakes
- map C-S1 through C-S12 into an unchanged conformance fixture and require migration, property, and crash evidence
- keep W3-W9 policy and every G4+ capability explicitly out of scope

Closes #627

### DIFF
--- a/docs/superpowers/plans/2026-08-05-psyche-w2-g2-foundation.md
+++ b/docs/superpowers/plans/2026-08-05-psyche-w2-g2-foundation.md
@@ -1,0 +1,1713 @@
+# Psyche W2 G2 Foundation Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Finish Psyche W2 and produce terminal G2 evidence for canonical versioned records, durable SQLite migrations, behavior-level fake boundaries, and migration/state-machine/property/crash tests.
+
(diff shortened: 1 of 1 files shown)


BunsDev:
Release review found four blockers before this plan is merge-ready:

1. **Append-only contradiction** (`:142`, `:1057-1065`, `:1707`): retention permits deleting non-latest transitions while the plan guarantees no API deletes historical transition rows. Exclude transitions from retention or explicitly revise the invariant first.
2. **Incomplete C-S3/C-S10 bindings** (`:1170-1206`, claim at `:1231-1235`): `SessionSnapshot`, `ResultBundle`, and `ArtifactReference` omit required graph/node/request-digest/familiar/project/lifetime correlations, so the frozen public behavior suite cannot prove mismatch rejection and complete artifact association required by `specs/psyche/COVEN_PREREQUISITES.md`.
3. **Invalid crash binary lookup** (`:1365-1377`): `cargo_bin!("crash_writer")` in a `psyche-store` integration test cannot resolve a binary target owned by `psyche-test-support`. Put the binary in the tested package or define an explicit cross-package build/lookup strategy.
4. **Unknown enum quarantine gap** (`:970-973`, `:1705`): `ingest` quarantines unknown majors/kinds only; typed decode failures from unknown enum values have no specified quarantine route despite the final requirement. Define and test that path.

Issue #627 is actively claimed by `buns@coven-g2-plan`, so I have not modified the branch.

BunsDev:
One internal contract mismatch needs correction before this plan becomes the implementation source of truth. The fixed `Store` boundary at the top declares `insert(&CanonicalDocument)`, and the later property test follows that contract, but Task 6’s primary immutability test calls `store.insert(&intent)` / `store.insert(&changed)` with typed `Intent` values. Wrap those fixtures in `CanonicalDocument::Intent(...)` (or deliberately revise the public boundary everywhere). As written, following the plan produces a compile failure unrelated to the intended RED step.

BunsDev:
Superseded by #643, reconstructed from current main as one DCO-clean commit after resolving all review and contract blockers. No history was force-pushed.

PR #631, opened 2026-08-05
fix: recover chaos event writes from SQLite locks

### LINKED ISSUE #615: benchmark-chaos fails on Linux CI: no output event, gating the perf job

BunsDev:
## Context

- Summary: Diagnoses the missing chaos-benchmark output as the event writer permanently latching a transient SQLite writer lock, fixes that recovery boundary, and retains bounded timeout evidence for future failures.
- Files changed: event-writer retry/tests, chaos benchmark diagnostics/tests, README, and implementation design/plan.
- Closes #615

## Implementation

- Approach: Keep the worker's current batch and retry only errors whose preserved `rusqlite::Error` is `DatabaseBusy` or `DatabaseLocked`. The policy allows four total attempts with 50 ms between attempts; exhausted lock errors and every unrelated error still use the existing permanent failure path.
- User-visible behavior: Advisory 1/8/32 chaos runs no longer lose all output after transient SQLite contention. If output still times out, the error includes bounded fixture-execution, session-event, and event-writer health evidence with fixture paths redacted.
- Compatibility notes: No API, schema, fixture concurrency, or benchmark timeout changes. The single worker still preserves batch ordering and waits to acknowledge critical events until commit.

## Verification

- [x] `cargo fmt --check`
- [x] `cargo test --workspace --locked`
- [x] Additional manual checks:
  - `cargo test -p coven-cli event_writer::tests::` (10 passed)
  - `node --test scripts/benchmark-cli.test.mjs scripts/benchmark-chaos.test.mjs` (56 passed)
  - Four consecutive real `benchmark-chaos.mjs` runs; every report passed `sessions_1`, `sessions_8`, and `sessions_32`
  - Independent code review found no significant issues

## Risk and Rollback

- Risk level: Medium-low. Lock failures can now delay a final writer failure for up to four SQLite busy-timeout attempts, but only while retaining the same ordered batch; arbitrary persistence errors remain fail-fast.
- Rollback plan: Revert commit `d02fdd6` to restore single-attempt writer behavior and the previous benchmark diagnostics.

## Agent Handoff

- Current state: Ready for CI review; local repository gates and repeated real collector runs pass.
- Follow-ups: Confirm the advisory Linux performance job produces its artifact on this PR.
- Known gaps: None beyond normal CI environment variance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -688,9 +688,10 @@ entries, so a trend artifact cannot be mistaken for a passing chaos test.
 CI runs this step as `continue-on-error`, so it collects trend data without
 gating merges — matching how the rest of this section describes these
 baselines. The deterministic fixture tests (`benchmark-chaos.test.mjs`) still
-gate. The 32-session scenario is intermittently unable to run on a two-core
-runner, where a live session can record no events at all; see #615. Expect
-`coven-chaos.json` to be absent from the uploaded artifact on those runs.
+gate. On timeout, the collector reports fixture execution count, session/event
+state, and a bounded event-writer health snapshot. The event writer retries
+only transient SQLite busy/locked commit failures before latching a permanent
+failure; unrelated persistence errors remain fail-fast.
 
 ### Architecture rules for contributors
 
--- a/crates/coven-cli/src/event_writer.rs
+++ b/crates/coven-cli/src/event_writer.rs
@@ -15,7 +15,7 @@ use std::{
 };
 
 use anyhow::{anyhow, Context, Result};
-use rusqlite::Connection;
+use rusqlite::{Connection, ErrorCode};
 use serde::Serialize;
 use serde_json::json;
 use uuid::Uuid;
@@ -27,6 +27,8 @@ const RESERVED_CRITICAL_BYTES: usize = 128 * 1024;
 const EVENT_OVERHEAD_BYTES: usize = 512;
 const MAX_BATCH_EVENTS: usize = 64;
 const COALESCE_WINDOW: Duration = Duration::from_millis(12);
+const SQLITE_LOCK_COMMIT_ATTEMPTS: usize = 4;
+const SQLITE_LOCK_RETRY_DELAY: Duration = Duration::from_millis(50);
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, serde::Deserialize)]
 #[serde(rename_all = "camelCase")]
@@ -303,7 +305,11 @@ fn run_worker(
     loop {
         let batch = take_batch(&shared);
         let bytes: usize = batch.iter().map(|item| item.bytes).sum();
-        let result = commit_batch(&mut conn, &coven_home, &batch);
+        let result = retry_transient_sqlite_lock(
+            || commit_batch(&mut conn, &coven_home, &batch),
+            SQLITE_LOCK_COMMIT_ATTEMPTS,
+            SQLITE_LOCK_RETRY_DELAY,
+        );
         match result {
             Ok(committed) => {
                 release_capacity(&shared, batch.len(), bytes);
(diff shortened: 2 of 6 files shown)


BunsDev:
Two blockers remain in the new timeout-diagnostics boundary:

1. Both `describeScenarioFailure` catch blocks pass `error.message` directly. JavaScript permits throwing non-`Error` values, including `null`/`undefined`; those catches can throw again and suppress the original timeout evidence. Normalize with something like `error instanceof Error ? error.message : String(error)` and add a test that injects a non-Error rejection into both request paths.
2. `redactedDiagnosticText` says/specifies that control characters are collapsed, but `/[\r\n\t]+/g` leaves NUL, ESC, and the rest of the C0/DEL range in logs. Collapse the full control range (preserving ordinary printable text) and pin it with a test.

The lock-only retry logic itself looks correct. Please also remove the duplicate `assert!(is_transient_sqlite_lock(&error));` in the persistent-lock test while updating the head.

BunsDev:
Superseded by https://github.com/OpenCoven/coven/pull/632, rebased safely onto current main without force-pushing, with DCO/public author metadata corrected and all review threads addressed.

PR #632, opened 2026-08-05
fix: recover event writer from transient locks

### LINKED ISSUE #615: benchmark-chaos fails on Linux CI: no output event, gating the perf job

BunsDev:
## Summary
- retry only transient SQLite busy/locked event-writer failures with a bounded backoff
- add bounded, redacted chaos timeout evidence for session state, child execution, and writer health
- collapse ASCII control characters and keep diagnostics resilient to non-Error throws

## Context
Clean replacement for conflicting PR #631, reconstructed on current `main` with DCO-compliant, public author metadata and all review feedback applied.

## Test Plan
- [x] cargo fmt --check
- [x] cargo clippy --workspace --all-targets -- -D warnings
- [x] cargo test --workspace --locked
- [x] node --test scripts/benchmark-chaos.test.mjs
- [x] python scripts/check-secrets.py
- [x] python3 scripts/check-coven-privacy.py --staged
- [x] git diff --check origin/main...HEAD

Closes #615

### DIFF
--- a/README.md
+++ b/README.md
@@ -688,9 +688,10 @@ entries, so a trend artifact cannot be mistaken for a passing chaos test.
 CI runs this step as `continue-on-error`, so it collects trend data without
 gating merges — matching how the rest of this section describes these
 baselines. The deterministic fixture tests (`benchmark-chaos.test.mjs`) still
-gate. The 32-session scenario is intermittently unable to run on a two-core
-runner, where a live session can record no events at all; see #615. Expect
-`coven-chaos.json` to be absent from the uploaded artifact on those runs.
+gate. On timeout, the collector reports fixture execution count, session/event
+state, and a bounded event-writer health snapshot. The event writer retries
+only transient SQLite busy/locked commit failures before latching a permanent
+failure; unrelated persistence errors remain fail-fast.
 
 ### Architecture rules for contributors
 
--- a/crates/coven-cli/src/event_writer.rs
+++ b/crates/coven-cli/src/event_writer.rs
@@ -15,7 +15,7 @@ use std::{
 };
 
 use anyhow::{anyhow, Context, Result};
-use rusqlite::Connection;
+use rusqlite::{Connection, ErrorCode};
 use serde::Serialize;
 use serde_json::json;
 use uuid::Uuid;
@@ -27,6 +27,8 @@ const RESERVED_CRITICAL_BYTES: usize = 128 * 1024;
 const EVENT_OVERHEAD_BYTES: usize = 512;
 const MAX_BATCH_EVENTS: usize = 64;
 const COALESCE_WINDOW: Duration = Duration::from_millis(12);
+const SQLITE_LOCK_COMMIT_ATTEMPTS: usize = 4;
+const SQLITE_LOCK_RETRY_DELAY: Duration = Duration::from_millis(50);
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, serde::Deserialize)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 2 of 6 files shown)


BunsDev:
One focused test correction before merge: the marker-order assertion can pass when `COVEN_BENCHMARK_MARKERS` is absent because `indexOf` returns `-1`, and `-1 < readyIndex` is true. Assert both tokens are present (or use a regex that requires marker write before ready output) before comparing their positions. The implementation is correct; this is the regression contract for the exact boundary being added.

BunsDev:
Marker-order regression contract tightened in 57d1f6c: both marker and ready tokens must be present before their positions are compared.

PR #639, opened 2026-08-05
docs: plan Psyche O1.1 conformance closure

BunsDev:
## Summary
- define the exact test matrix needed to close the merged O1.1 corrective annex
- preserve the boundary against later Psyche adoption, fencing, cancellation-acknowledgement, and dispatch contracts
- require signed commits and durable post-merge evidence in GitHub and Beads

## Validation
- `python3 scripts/check-api-contract-docs-test.py`
- `python3 scripts/check-api-contract-docs.py`
- `git diff --check origin/main...HEAD`

### DIFF
--- a/docs/superpowers/plans/2026-08-05-psyche-o1-1-conformance-closure.md
+++ b/docs/superpowers/plans/2026-08-05-psyche-o1-1-conformance-closure.md
@@ -71,7 +71,7 @@ printf '%s\n' '{"type":"turn.completed"}'
 }
 ```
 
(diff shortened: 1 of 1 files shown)


BunsDev:
Blocking plan gaps: (1) the continuation matrix only uses ledger IDs and never verifies the required stable conversation-key path through sibling creation and native harness resume; (2) the malformed-stream fixture has no descendant, so it cannot prove malformed output terminates and reaps descendants; (3) the claimed signal restoration coverage never checks that the calling thread signal mask is restored after finish. Please add explicit tests for all three contracts before implementation begins.

PR #640, opened 2026-08-05
fix(bench): report sampled runtime metrics accurately

BunsDev:
## Summary
- close the throughput timing window before observer shutdown
- label periodic writer queue observations as sampled maxima and bump the report schema
- describe cancellation as cancellation-to-terminal latency rather than acknowledgement

## Validation
- `node --test scripts/benchmark-chaos.test.mjs`
- `cargo build -p coven-cli --locked --quiet`
- `node scripts/benchmark-chaos.mjs --binary target/debug/coven --output /tmp/coven-chaos-metric-semantics.json` (1/8/32 scenarios passed)

### DIFF
--- a/README.md
+++ b/README.md
@@ -677,11 +677,13 @@ node scripts/benchmark-chaos.mjs --binary target/debug/coven --output /tmp/coven
 
 The command always exercises 1, 8, and 32 concurrent deterministic harness
 sessions, records launch-to-first-output percentiles, throughput, cancellation
-acknowledgement, SQLite file growth, writer connection/transaction deltas, peak
-writer backlog, and sampled daemon RSS, and writes a redacted JSON report. The
-writer measurements come from the live health contract; RSS is resolved by the
-exact daemon PID through `coven pc top --json`, without retaining process names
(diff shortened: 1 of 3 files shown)


BunsDev:
Two measurement-contract blockers remain: (1) periodic writer observation stops before cancellation and only one post-cancellation snapshot is taken, so transient cancellation backlog can drain unseen while maxSampledEvents/maxSampledBytes are reported as maxima; keep sampling through cancellation and the terminal snapshot. (2) diskFull is still marked covered by a low-space test that explicitly performs no database write, so it does not exercise SQLITE_FULL, writer failure, or recovery; mark this blocked/partial until a real write-fault seam is tested.

BunsDev:
The new head still leaves both measurement blockers unchanged: periodic writer sampling stops before cancellation and a single terminal snapshot can miss transient cancellation backlog; diskFull is still reported covered by a low-space test that never opens or writes SQLite, so it does not exercise SQLITE_FULL, writer failure, or recovery. Please keep sampling through cancellation and either add a deterministic write-failure/recovery seam or report disk-full as unavailable.

PR #645, opened 2026-08-05
fix(health): keep status reads side-effect free

BunsDev:
## Summary
- keep `/health` and `/hub/status` on read-only existing-store connections so status reads cannot create SQLite files or mutate metadata
- initialize and backfill the stable hub identity at writable daemon startup and node registration boundaries
- fail closed on malformed privacy configuration before manual log pruning can open or delete from the store

## Validation
- `cargo test -p coven-cli hub --no-fail-fast` (34 passed)
- `cargo test -p coven-cli daemon_store_initialization_prepares_request_connections --no-fail-fast`
- `cargo test -p coven-cli routes_health_request_to_json --no-fail-fast`
- `cargo clippy -p coven-cli --all-targets -- -D warnings`
- `git diff --check`

Follow-up to #637.

### DIFF
--- a/crates/coven-cli/src/api.rs
+++ b/crates/coven-cli/src/api.rs
@@ -6766,7 +6766,7 @@ mod tests {
         assert!(response.body.contains(r#""scheduler":true"#));
         assert!(response.body.contains(r#""hub":true"#));
         assert!(response.body.contains(r#""executorDispatch":true"#));
-        assert!(response.body.contains(r#""role":"hub""#));
+        assert!(!response.body.contains(r#""role":"hub""#));
         assert!(response.body.contains(r#""structuredErrors":true"#));
         assert!(response.body.contains(r#""storage":{"status""#));
         assert!(response.body.contains(r#""writerBacklogEvents":0"#));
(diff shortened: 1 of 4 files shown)


BunsDev:
Blocking side-effect gap: health and offline hub-status reads still open an existing WAL-mode database through SQLite. That open can create or update coven.sqlite3-shm / -wal sidecars, and can fail when the store directory is non-writable. The new test only covers a missing database, so it does not prove existing-store status reads are side-effect free. Please add existing WAL-store coverage in a non-writable directory and use a genuinely immutable/read-only path that cannot create sidecars.

BunsDev:
The new head still does not resolve the side-effect contract. SQLITE_OPEN_READ_ONLY alone can create a -shm sidecar for an existing WAL database, or fail when the directory is non-writable; the new test still covers only a missing database. Add existing WAL-store tests with absent sidecars in writable and non-writable directories and use a genuinely side-effect-free WAL-compatible read strategy. Also, required Ubuntu CI is PR-caused red because cargo fmt --check fails in the new api.rs test edit. The benchmark failure occurred before measurement in an external-session fixture and is not evidence of a performance regression.

BunsDev:
Superseded by #646, reconstructed from current main as a single signed commit with preserved contributor attribution. Closing this head because GitHub Autofix added an unsigned commit using a private email.

PR #646, opened 2026-08-05
fix(health): keep status reads side-effect free

BunsDev:
## Summary
- serve `/health` and `/hub/status` from daemon-owned, versioned in-memory snapshots so GET requests never open or mutate SQLite/WAL/SHM state
- initialize and backfill stable hub identity at writable startup/registration boundaries, refresh snapshots after mutations, and fail unavailable instead of serving stale state
- cache storage health at writable startup/maintenance boundaries and degrade safely when privacy configuration or refresh fails
- fail closed on malformed privacy configuration before manual log pruning can open or delete from the store

## Validation
- `cargo fmt --check`
- `cargo test --workspace --locked` (1,810 unit tests pass; one unrelated daemon-start integration race reproduces unchanged on `main`)
- focused health, hub, storage-health, daemon-startup, WAL side-effect, concurrency, and error-path regressions
- `git diff --check`

Follow-up to #637. Supersedes #645, whose head was contaminated by an unsigned autofix commit with unsuitable attribution.

### DIFF
--- a/crates/coven-cli/src/api.rs
+++ b/crates/coven-cli/src/api.rs
@@ -289,12 +289,19 @@ fn health_response_with_hub(
         response.hub = serde_json::from_value(summary).ok();
     }
     response.storage = Some(
-        store::storage_health(coven_home, event_writer.as_ref()).unwrap_or_else(|error| {
+        store::cached_storage_health(coven_home, event_writer.as_ref()).unwrap_or_else(|error| {
             store::unavailable_storage_health(coven_home, error, None, event_writer.as_ref())
         }),
     );
     response.event_writer = event_writer;
     response
+}
+
+fn hub_mutation_response(
+    coven_home: &Path,
+    operation: impl FnOnce() -> Result<ApiResponse>,
+) -> Result<ApiResponse> {
+    crate::hub::with_status_mutation(coven_home, operation)
 }
 
 #[allow(dead_code)]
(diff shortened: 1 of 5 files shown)


BunsDev:
The replacement still does not satisfy the side-effect-free contract. Health and hub status continue to use SQLITE_OPEN_READ_ONLY, which may create -shm/-wal sidecars for an existing WAL database or fail in a non-writable directory. No immutable/snapshot strategy was added, and the tests cover empty homes or database length rather than absent sidecars plus non-writable access. Please test an existing WAL store with removed sidecars in writable and non-writable directories, assert no filesystem changes, and still return current WAL data.

BunsDev:
Resolved the WAL side-effect blocker in signed commit `a107950`:

- `/health` and `/hub/status` now read daemon-owned in-memory snapshots and never open SQLite on GET
- versioned publication prevents stale concurrent refreshes; refresh failures invalidate status without turning a committed mutation into a retryable failure
- storage health is refreshed only at writable daemon startup/maintenance boundaries and degrades safely on malformed privacy config
- added initialized-store, absent-sidecar, non-writable-directory, failed-dispatch, partial-write, stale-publication, and malformed-config regressions

Local gates passed: formatting, workspace clippy with `-D warnings`, 1,810 unit tests, task-specific integration tests, secret scan, privacy guard, and diff checks. The full workspace integration pass reaches one pre-existing `concurrent_daemon_start_commands_share_one_daemon` race that reproduces unchanged on `main`.

BunsDev:
The new snapshot approach resolves the side-effect-free WAL read blocker, but introduces a low-disk regression in daemon.rs: after run_scheduled_maintenance returns blocked_by_free_disk, the scheduler still immediately calls open_initialized_store to refresh the snapshot. That defeats the no-store-access safety path and can create DB/WAL sidecars or fail on exhausted/non-writable storage. Preserve the maintenance report and skip connection-based snapshot refresh when blocked_by_free_disk is true; add scheduler-level coverage for this path.

BunsDev:
Addressed the low-disk scheduler regression in signed commit `eb5b8a7`:

- the scheduler now preserves the `ScheduledMaintenanceReport`
- `blocked_by_free_disk` returns before any SQLite-backed snapshot refresh
- a daemon scheduler-level regression proves blocked maintenance creates no DB/WAL/SHM files

Exact-head local gates pass: `cargo fmt --check`, workspace clippy with `-D warnings`, `cargo test --workspace --locked` (1,811 unit tests plus integrations), secret scan, staged privacy guard, and diff checks.

BunsDev:
The scheduler-level low-disk reopen regression is fixed, but one health-priority blocker remains: store.rs checks snapshot.degraded before live maintenance_blocked. If a snapshot was already degraded and disk later falls below the critical watermark, health stays degraded instead of the documented critical state for the entire blocked period. Give live low-disk status precedence while retaining the prior lastMaintenanceError.

PR #600, opened 2026-08-04
docs(specs): native Agent File System (AFS) research brief

BunsDev:
Adds `specs/coven-agent-fs/RESEARCH.md` — the deep-research brief on building a native OpenCoven Agent File System, modeled on (but independent of) Turso AgentFS.

**Recommendation:** build a native `coven-afs` crate in the daemon — AgentFS SPEC v0.4-compatible SQLite schema (inode FS + CoW overlay with whiteouts/fs_origin + insert-only audit + KV), coven provenance extension tables (session/familiar/bead), `fuser`/`nfsserve` mounts, delta-to-git commit materialization.

Includes per-RQ findings with confidence labels, a design-family comparison table, open questions, and a 16-source ledger.

Research artifact only — no code changes.

PR #658, opened 2026-08-06
feat: coven-afs crate — AgentFS SPEC v0.4 schema + overlay spike

BunsDev:
Implements bead **coven-d4p** — the core storage spike from [`specs/coven-agent-fs/RESEARCH.md`](https://github.com/OpenCoven/coven/blob/main/specs/coven-agent-fs/RESEARCH.md) (PR #600).

## What

New `crates/coven-afs`: an [AgentFS SPEC v0.4](https://github.com/tursodatabase/agentfs/blob/main/SPEC.md)-compatible SQLite storage engine.

- **Schema verbatim from SPEC v0.4**: `fs_config` (immutable `chunk_size`, default 4096), `fs_inode` (nsec columns + rdev), `fs_dentry` (`UNIQUE(parent_ino,name)`, hard-link capable), `fs_data` chunked BLOBs, `fs_symlink`, `fs_whiteout` (parent_path-indexed), `fs_origin`, `kv_store` (JSON), insert-only `tool_calls`.
- **Research conflict C1 resolved**: `fs_data` confirmed authoritative — it appears in upstream SPEC.md and all four SDKs (Rust/Go/Python/TS); `fs_block` has zero hits in upstream code. Frozen accordingly (noted in `schema.rs`).
- **Core ops**: path resolution, chunked write/read + offset reads, `mkdir_p`, readdir, stat, rename, delete (nlink-aware), hard links, symlinks.
- **Overlay**: writable delta over read-only base; lookup order **delta → whiteout → base**; whiteout lifecycle (created only when shadowing base, removed on re-creation); full-file copy-up preserving metadata; `fs_origin` keeps base inode numbers visible through stat (kernel-cache safety).
- **Audit + KV**: tool-call mutual-exclusion and `duration_ms` rules enforced at the API; no update/delete surface (insert-only by construction).
- **`check_consistency()`** verifies the SPEC's 8 filesystem rules plus chunk-shape invariants — used as an oracle by the tests.

## Not in scope

No mount layer (FUSE/nfsserve) — that's bead **coven-110**, blocked on this.

## Verification

- `cargo test --workspace --locked` — all green (24 new integration tests + 4 path unit tests; 1809 existing coven-cli tests unaffected)
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo fmt --check` — clean

Spec compatibility is deliberate: DBs produced by this crate are readable by upstream `agentfs` tooling and vice versa.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -708,6 +708,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "773648b94d0e5d620f64f280777445740e61fe701025087ec8b57f45c791888b"
 
 [[package]]
+name = "coven-afs"
+version = "0.1.0"
+dependencies = [
+ "rusqlite",
+ "serde",
+ "serde_json",
+ "tempfile",
+ "thiserror 2.0.18",
+]
+
+[[package]]
 name = "coven-cli"
 version = "0.0.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = ["crates/coven-cli", "crates/coven-memory", "crates/coven-relay"]
+members = ["crates/coven-afs", "crates/coven-cli", "crates/coven-memory", "crates/coven-relay"]
 resolver = "2"
 
 [workspace.package]
--- a/crates/coven-afs/Cargo.toml
+++ b/crates/coven-afs/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "coven-afs"
+version = "0.1.0"
+description = "Native OpenCoven agent filesystem: AgentFS SPEC v0.4-compatible SQLite storage with copy-on-write overlay"
+edition.workspace = true
+license.workspace = true
+authors.workspace = true
+repository.workspace = true
+
+[dependencies]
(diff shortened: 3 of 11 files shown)


BunsDev:
Blocking overlay defects found during source review:

1. A delta regular file can shadow a base directory while `readdir` and descendant reads still fall through to the base directory. In `overlay.rs`, `write_file("/docs", ...)` accepts a base `/docs` directory (lines 167-188), then `readdir("/docs")` independently accepts the base directory when the delta entry is not a directory (251-280). Resolve each component through the merged namespace and require every parent to be a directory; add a regression covering a file-over-directory shadow.

2. `readdir` performs linear `Vec::contains` checks for every base entry against whiteouts and accumulated names (262-280), producing quadratic behavior for base directories. Merge with `HashSet`s and sort the final names once; add a large-directory regression or focused complexity-safe coverage.

This should be corrected before merge.

BunsDev:
Returning this to draft as an explicit spike. Blocking correctness gaps remain in ancestor shadowing/whiteout resolution, merged inode uniqueness, and atomic rename replacement; writable-base enforcement and origin cleanup also need work. Rebase and fresh CI are required after those semantics are fixed.

PR #661, opened 2026-08-07
fix(session): bound handoff cursor read and flush truncation across direct events (#613, #642)

### LINKED ISSUE #613: Session handoff: unbounded event scan for the cursor, and cursor equality blocks live claims
Follow-up from #606, which merged at 2181e0f. Both findings were raised on the PR before it merged and are unresolved on `main`.

## 1. Reading the cursor scans the session's entire event history

Three handlers in `crates/coven-cli/src/api.rs` (lines 2012, 2087, 2151 on `main`) read the latest event sequence like this:

```rust
let event_cursor = store::list_events(&conn, session_id)?
    .last()
    .map(|event| event.seq)
    .unwrap_or(0);
```

`store::list_events` delegates to `list_events_with_options` with `EventsQueryOptions::default()` — **no limit**. Every call materializes every event row for the session into a `Vec<EventRecord>`, including full raw PTY output payloads, in order to read one integer off the end.

The cost grows without bound over a session's life, and it is paid on every handoff emit, claim, and acknowledgement.

This got materially worse in the same batch: #607 (bounded event writer, merged at 114f9c8) exists specifically to raise sustained event throughput. The two landed within minutes of each other, so the volume this scan walks is now higher by design.

`SELECT MAX(seq) FROM events WHERE session_id = ?1` answers the question directly. `list_events_with_options` is already there if an actual record is needed.

## 2. Cursor equality makes a live-session handoff unclaimable

`emit_handoff` requires the session to be live:

```rust
if session.status != "running" {
    return session_not_live_response(session_id);
}
```

and both `claim_session_handoff` and `acknowledge_session_handoff` reject on **any** cursor movement:

```rust
if current_cursor != handoff.event_cursor {
    return api_error(409, "transcript_diverged", ...);
}
```

Together these say: the handoff is claimable only if the session emitted zero events between emit and claim — but the session is required to be `running`, and a running harness session emitting output is the normal case. Any output in that window 409s the claim permanently. The handoff record can never be claimed, only superseded by a fresh one that races identically.

Two coherent resolutions:

- **If the intent is "my snapshot is still a prefix of the current transcript"** — then divergence is `current_cursor < handoff.event_cursor` (truncated or rewritten), and forward movement should be allowed, with the claimant resuming from `handoff.event_cursor`.
- **If strict equality is genuinely intended** — then handoff only works on quiesced sessions, and `emit_handoff` should require that rather than requiring `running`.

As merged, I don't see the path where a busy session's handoff gets claimed. Happy to be corrected on the intent.

Context: [review on #606](https://github.com/OpenCoven/coven/pull/606#issuecomment-5187620299).

### LINKED ISSUE #642: fix: preserve truncation boundaries across direct session events
## Problem\n\nPR #638 added per-session output truncation markers, but accepted input, kill, and targeted cast events still persist directly through Store::insert_event rather than EventWriter. A pending truncation episode can therefore be emitted after an event that should close the gap, and later drops can combine across that event.\n\n## Required correction\n\n- Route all accepted live-session events through the shared EventWriter, or synchronously flush the session's pending truncation marker before direct insertion.\n- Add integration tests for drop -> input/kill/cast -> drop -> recovered output.\n- Assert the marker precedes each boundary event and separate pressure episodes produce separate markers.\n\n## Context\n\nTracked from the blocking review on #638, which merged at e9118f078cd635996585fd7a5251f17758b0a084 despite the finding.

CompleteDotTech:
Closes #613. Closes #642.

## #613(A) — bound the handoff cursor read
`emit_handoff` / `claim_session_handoff` / `acknowledge_session_handoff` read the latest sequence via `store::list_events(...).last()`, materializing **every** event row (incl. raw PTY payloads) to read one integer — an unbounded cost paid on every handoff. Added `store::max_event_seq` (`SELECT MAX(seq)`) and replaced the three full scans. Semantically equivalent to the old `.last().seq`.

## #613(B) — allow forward-cursor claims (prefix / forward-allowed contract)
As merged, a `running` session that emitted **any** output between handoff emit and claim 409'd permanently (`current_cursor != handoff.event_cursor`), so a busy session's handoff was never claimable.

- Divergence is now only `current_cursor < handoff.event_cursor` (transcript truncated/rewritten below the snapshot). Forward movement is expected and allowed.
- The claim response surfaces `resumeFrom = handoff.event_cursor`, so the claimant replays exactly the recorded prefix rather than the possibly-advanced live tail.
- `acknowledge_session_handoff` acknowledges against the snapshot cursor (not the live tail) to keep the store-level consistency invariant.
- `emit_handoff` still requires `running` (unchanged).
- Intent chosen deliberately (prefix over strict-quiesced) — see the discussion on #613.

## #642 — flush truncation markers across direct session events
Accepted input/kill/targeted-cast events bypassed the shared `EventWriter`, so a pending truncation marker could be emitted *after* the event that should close the gap. Now routed through `EventWriter` so the marker flushes first; adds ordering + separate-episode coverage.

## Tests
- `handoff_claim_allows_forward_cursor_and_exposes_resume_from` (new) — forward cursor claims succeed and expose `resumeFrom` (red→green: 409→200).
- `handoff_claim_fails_closed_when_transcript_or_workspace_diverges` (updated) — now simulates truncation (backward cursor) + preserves workspace-divergence coverage, still 409.
- `#642` marker-ordering / separate-episode tests.

## Readiness
- `cargo fmt --check` — clean · `cargo clippy -p coven-cli --all-targets -- -D warnings` — clean
- `cargo test -p coven-cli` — all handoff/api tests green (16/16 handoff). The 65 `memory_import` failures are pre-existing WSL `/mnt/c` filesystem-semantics artifacts, confirmed identical on clean `HEAD` via `git stash` — not introduced here; CI runs on Linux.
- secret + privacy scans — clean. Branch based on `origin/main` (`f3d9585`). Commits SSH-signed.

### DIFF
--- a/crates/coven-cli/src/api.rs
+++ b/crates/coven-cli/src/api.rs
@@ -2132,11 +2132,16 @@ fn claim_session_handoff(coven_home: &Path, path: &str, body: Option<&str>) -> R
         .context("stored handoff workspace snapshot is invalid")?;
     let current_workspace = WorkspaceSnapshot::capture(Path::new(&session.project_root));
     let current_cursor = store::max_event_seq(&conn, session_id)?;
-    if current_cursor != handoff.event_cursor {
+    // Prefix / forward-allowed contract: the handoff snapshot describes a prefix
+    // of the live transcript. Forward movement (the running session emitting more
+    // output after the snapshot) is expected and NOT divergence. Only a backward
+    // cursor -- the transcript being truncated or rewritten below the snapshot
+    // point -- means the recorded prefix no longer exists, so we fail closed.
+    if current_cursor < handoff.event_cursor {
         return api_error(
             409,
             "transcript_diverged",
-            "Source transcript changed after the handoff snapshot.",
+            "Source transcript was truncated below the handoff snapshot.",
             Some(
                 json!({ "handoffId": handoff_id, "expectedCursor": handoff.event_cursor, "actualCursor": current_cursor }),
             ),
@@ -2165,7 +2170,13 @@ fn claim_session_handoff(coven_home: &Path, path: &str, body: Option<&str>) -> R
     };
     json_response(
         200,
-        &json!({ "handoff": claimed, "sourceInputFenced": true }),
+        &json!({
+            "handoff": claimed,
+            "sourceInputFenced": true,
+            // The claimant must resume from the snapshot point, not the current
+            // (possibly-advanced) tail, so it replays exactly the recorded prefix.
+            "resumeFrom": handoff.event_cursor,
+        }),
     )
 }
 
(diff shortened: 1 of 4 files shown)


BunsDev:
PR #667 has merged #613 into main, so this branch now conflicts by carrying the same handoff-cursor work. Please rebase onto current main, drop the #613 changes, retain only the direct-event truncation fix for #642, and update the PR body to close #642 only. Do not merge this combined branch as-is.

BunsDev:
Superseded by #669, the current-main #642-only repair. #669 preserves the reviewed #613 retention and atomicity fixes already merged in #667, so this combined conflicting branch must not be merged.

Issue #648, opened 2026-08-05
fix: make cold daemon startup metrics readiness-bound and statistically valid


BunsDev:
## Problem

PR #647 merged with two unresolved measurement-contract defects.

## Required corrections

- Measure cold startup only until the first valid health response. The current synchronous daemon-start timing includes post-readiness duplicate-daemon cleanup and a full process-table scan, so results vary with unrelated runner state. Await command completion and cleanup outside the timed interval.
- Do not report p95/p99 from only three CI samples; both nearest-rank values collapse to the single maximum. Increase the cold-start-specific sample count substantially or omit unsupported tail percentiles.

## Context

Tracked from the blocking review on #647, merged unchanged at head b34a552c53140162b2e39e8b12a4080435724e9d as commit 0832ac73c8e7d83ff276a003a8d9795a504dc67f.

## Program tracking
- Bead: `cave-3qas7.6.1`
- Familiar owner: Sage
- Parent program: OpenCoven/coven#599
- Cave Board and GitHub Project visibility are linked from the Bead.

Issue #636, opened 2026-08-05
fix(perf): correct runtime baseline measurement semantics


BunsDev:
PR #635 merged with three unresolved measurement-contract findings:

1. `scripts/benchmark-chaos.mjs` captures `completedAt` after awaiting the observer loop, so throughput includes up to one sampling interval plus in-flight health-request latency. Capture workload completion immediately when launches resolve.
2. `peakEvents` / `peakBytes` are 25 ms sampled maxima, not true event-writer high-watermarks, and observation stops before cancellation. Rename/document them as sampled maxima and sample the complete measured interval, or add daemon-maintained high-water counters.
3. `diskFull` is marked covered by `scheduled_maintenance_below_watermark_does_not_open_or_write_the_store`, which proves low-disk maintenance gating but never exercises `SQLITE_FULL`, write failure handling, or recovery. Restore blocked/partial status until an actual write-fault seam is tested.

The PR and post-merge `main` CI are green; these are semantic validity defects rather than compilation failures. See review comments on #635.

## Program tracking
- Bead: `cave-3qas7.1.1`
- Familiar owner: Sage
- Parent program: OpenCoven/coven#599
- Cave Board and GitHub Project visibility are linked from the Bead.

BunsDev:
PR #640 merged as 40a39a6fc23c9fdda0bcf682d3e7c9da36a725bf, but re-review at head cb99e11255d85e3841044c2cb4db7a3dead00a17 found both tracked defects still present: sampling stops before cancellation, and disk-full remains claimed by a low-space no-write test. This issue remains required corrective work.

PR #662, opened 2026-08-07
fix(bench): correct runtime baseline and cold-start metric semantics (#636, #648)

CompleteDotTech:
Refs #636. Refs #648.

Corrects the completed-window, sampled-maxima, and small-sample percentile semantics. Deliberately leaves disk-full coverage blocked until a real `SQLITE_FULL` write-failure and recovery regression lands.

## Verification
- exact-head CI: all required checks green
- `node --test scripts/benchmark-cli.test.mjs scripts/benchmark-chaos.test.mjs` (65/65)
- `python scripts/check-secrets.py`
- `git diff ... --check`

### DIFF
--- a/scripts/benchmark-chaos.mjs
+++ b/scripts/benchmark-chaos.mjs
@@ -110,7 +110,9 @@ export function storageMetricStatus() {
     },
     eventQueueDepth: {
       status: 'measured',
-      source: 'maxima of periodic eventWriter.queuedEvents/eventWriter.queuedBytes samples'
+      source:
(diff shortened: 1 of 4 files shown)


BunsDev:
Blocking measurement-boundary issue: `measureColdDaemonStarts` still times the complete synchronous `startDaemon()` call. The CLI daemon-start path waits for readiness and then performs duplicate-daemon cleanup/process-table refresh, so the sample includes post-readiness work. Start the daemon asynchronously, poll the local health endpoint independently, record elapsed time at the first successful health response, then await starter completion/cleanup outside the measured interval. Add a test that distinguishes readiness from later starter completion.

PR #669, opened 2026-08-08
fix(session): preserve direct event truncation boundaries (#642)

### LINKED ISSUE #642: fix: preserve truncation boundaries across direct session events

BunsDev:
## Summary
- route accepted live input, kill, and targeted cast events through the daemon EventWriter
- preserve separate output-truncation episodes at each direct-event boundary
- reject oversized writer-backed inputs and targeted casts before side effects while preserving direct cockpit casts

## Test Plan
- [x] cargo fmt --check
- [x] cargo clippy --workspace --all-targets -- -D warnings
- [x] cargo test --workspace --locked
- [x] python scripts/check-secrets.py
- [x] python3 scripts/check-coven-privacy.py --staged

Closes #642
Supersedes #661

### DIFF
--- a/crates/coven-cli/src/api.rs
+++ b/crates/coven-cli/src/api.rs
@@ -235,6 +235,15 @@ pub trait SessionRuntime {
     fn send_input(&self, session_id: &str, payload: &Value) -> Result<()>;
     fn kill_session(&self, session_id: &str) -> Result<()>;
 
+    fn record_session_event(
+        &self,
+        _session_id: &str,
+        _kind: &str,
+        _payload: &Value,
+    ) -> Option<Result<()>> {
+        None
+    }
+
     fn event_writer_health(&self) -> Option<crate::event_writer::EventWriterHealth> {
         None
     }
@@ -377,7 +386,7 @@ pub fn handle_request_with_runtime(
             let (status, response) = control_plane::route_action(payload);
             json_response(status, &response)
         }
-        ("POST", "/cast") => submit_cast(coven_home, body),
+        ("POST", "/cast") => submit_cast(coven_home, body, runtime),
         ("GET", "/cast-codes") => cast_codes_response(),
         // Filesystem-backed reads under ~/.coven/. Missing files return [].
         ("GET", "/familiars") => {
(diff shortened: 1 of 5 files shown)


BunsDev:
Addressed the concurrency review blocker:

- input now reserves the writer boundary before transport I/O and restores the episode on failure;
- concurrent kill is still issued while input is blocked;
- kill and exit share a per-registration ordering guard so durable order is `kill -> exit`;
- added deterministic reservation, rollback, blocked-input, and exit-race regressions.

Full workspace tests, clippy, formatting, secret/privacy guards, and final review passed.

BunsDev:
Superseded by #679, which is merged.

Verified against current `main` (`d6ba6b9`): every change on this branch is already present.

- `crates/coven-cli/src/api.rs` — identical blob to `main`
- `crates/coven-cli/src/daemon.rs` — identical blob to `main`
- `docs/superpowers/plans/2026-08-07-direct-event-truncation-boundaries.md` — identical blob
- `docs/superpowers/specs/2026-08-07-direct-event-truncation-boundaries-design.md` — identical blob
- `crates/coven-cli/src/event_writer.rs` — the only file that differs, and `main` is a strict superset: it carries all 168 of this branch's added lines (including the `pub(crate) fn start_with_capacity` change this PR required) plus #676's SQLITE_FULL classification, setup injection, and recovery regression test.

The conflict here is purely that this branch predates #676. Merging it would resolve to `main`'s content anyway, so there is nothing left to land. #642 is closed by #679.

BunsDev:
Closing as superseded by #679 (merged). All changes verified present on main.

PR #684, opened 2026-08-09
feat(afs): serve AFS sessions, provenance, diff, and timeline from the daemon

BunsDev:
## Context

- Summary: implements the storage and read surfaces of `specs/coven-agent-fs/DESIGN.md` §3–4 for bead `coven-5kt`, now that the engine (`coven-d4p`) and the mount spike (`coven-110`) have landed.
- Files changed: `crates/coven-afs/src/{provenance,diff,fs,ino,lib}.rs`, `crates/coven-afs/tests/spec_consistency.rs`, `crates/coven-cli/src/{afs,api,main}.rs`, `crates/coven-cli/Cargo.toml`, `docs/daemon/socket-api.md`
- Closes: bead `coven-5kt` **partially** — see *Scope* below.

## Implementation

**Extension tables** (`afs_session`, `afs_provenance`, `afs_commit`) created lazily with every column nullable or defaulted, so a database produced by upstream `agentfs` reads as one with no provenance rather than erroring (rule **E3**). Nothing writes to a SPEC table (**E1**).

**Rule E2 is tested, not asserted.** `drop_extensions()` plus a conformance test that runs the SPEC consistency rules, drops every `afs_*` table, and runs them again — verifying file contents, symlinks, and directories are unchanged and that the coven accessors still degrade cleanly afterwards.

**Diff is computed from `fs_dentry` / `fs_origin` / `fs_whiteout`, never from the provenance log** (§4.4). Writes can arrive with no bound actor, so a provenance-derived diff would silently omit exactly the changes nobody can account for. There is a test for precisely that: a write with no provenance row still appears, marked `attribution: "unknown"`.

**Findings from the spike carried forward.** WAL is enabled on every delta open — the coven-110 numbers put whole-file writes at 8.84x the host filesystem without it and 1.29x with it. `COPY_UP_MAX_BYTES` is 16 MiB, justified from the measured copy-up curve (1 MiB → 3.3 ms, 8 MiB → 57 ms, 32 MiB → 238 ms), keeping a worst-case first write near ~120 ms.

Two smaller decisions worth flagging:

- `checkpoint_to_single_file()` was added because publishing a WAL base left `-wal`/`-shm` sidecars in `bases/`, breaking the single-portable-file property a shared base depends on. A test caught it.
- Oversized and excluded files during ingest are **skipped and counted**, not silently dropped — a base that quietly omits files would make a diff look complete when it is not.

## Scope — what is deferred, and why

Seven of nine operations ship: `create`, `list`, `get`, `join`, `diff`, `timeline`, `discard`.

**`commit` and `mount` are deferred.** They return `501` with `afs.commit_unsupported` / `afs.mount_unsupported`, and `health` advertises `afsCommit:false` / `afsMount:false`, so a client learns this from the capability handshake rather than from a failed request. Reasons: mount is blocked on `coven-x77` (an agent process could not write through the NFS mount on macOS), and commit materialization is a git-worktree-plus-signing, all-or-nothing operation that deserves its own change rather than being bolted onto this one. I have not filed the commit bead — say the word and I will, or push it into this PR if you would rather it land together.

Consistent with that, the error enum defines only codes the implemented operations can raise. `afs.copy_up_too_large`, `afs.path_outside_root`, `afs.base_diverged`, and `afs.commit_conflict` arrive with commit; defining them now would advertise a contract the daemon does not honour. Clippy caught two of them sitting unconstructed, which is what prompted the cleanup.

## Verification

- [x] `cargo fmt --check`
- [x] `cargo test --workspace --locked` — 1848 passed, 2 ignored (was 1833)
- [x] `python3 scripts/check-secrets.py` — clean
- [x] Additional manual checks: `python3 scripts/check-api-contract-docs.py` passed; privacy guard passed on 12 staged files; `git diff --check` clean.

15 new tests: 5 provenance (including a foreign database with no extensions reading as unbound), 4 diff (including the unattributed-write case and directory handling), 1 E1–E3 conformance, 11 store-level, 4 route-level covering the capability handshake, structured errors, discard confirmation, and timeline pagination.

## Risk and Rollback

- Risk level: low. Additive routes behind a capability flag; no existing route, schema, or contract changes. `coven-afs` gains no default-feature dependencies.
- Rollback plan: revert the commit.

## Agent Handoff

- Current state: the read and lifecycle surface is complete and tested end to end through `handle_request`.
- Follow-ups: commit materialization (DESIGN.md §5); mount routes once `coven-x77` resolves; `coven-gr1` (Cave surfaces) is unblocked by this and can consume these routes.
- Known gaps: diff has no per-path unified-diff mode yet (`truncated` is always `false` — the field is in the shape so adding it later is additive); the store opens databases per request rather than pooling, which is right for correctness but unmeasured under concurrent load.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -747,6 +747,7 @@ dependencies = [
  "chrono",
  "clap",
  "clap_complete",
+ "coven-afs",
  "coven-runtime-spec",
  "coven-threads-core",
  "crossterm",
--- a/crates/coven-afs/src/diff.rs
+++ b/crates/coven-afs/src/diff.rs
@@ -1,0 +1,237 @@
+//! Change set between an overlay's delta and its read-only base.
+//!
+//! Computed from `fs_dentry`, `fs_origin`, and `fs_whiteout` — never from
+//! `afs_provenance`. DESIGN.md §4.4 is explicit about why: writes can arrive
+//! without a bound actor (upstream `agentfs` tooling, a manual `sqlite3` edit,
+//! a mount write from a process the daemon did not launch), and those produce
+//! no provenance row. The filesystem tables cannot lie about what the
+//! filesystem contains; the provenance log can be incomplete. A diff derived
+//! from provenance would silently omit exactly the changes nobody can account
+//! for, which are the ones worth seeing.
+
+use crate::fs::Metadata;
+use crate::{AgentFs, OverlayFs, Result};
+
+/// How a path differs from the base.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Change {
+    Added,
+    Modified,
+    Deleted,
+}
+
+impl Change {
+    pub fn as_str(self) -> &'static str {
+        match self {
+            Self::Added => "added",
+            Self::Modified => "modified",
+            Self::Deleted => "deleted",
+        }
+    }
+}
+
+/// One changed path.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ChangeEntry {
+    pub path: String,
+    pub change: Change,
+    pub bytes: i64,
+    pub ino: Option<i64>,
+    pub base_ino: Option<i64>,
+    pub mode: Option<u32>,
+}
+
+/// The full change set, with counts callers would otherwise recompute.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct ChangeSet {
+    pub entries: Vec<ChangeEntry>,
+    pub added: usize,
+    pub modified: usize,
+    pub deleted: usize,
+    pub bytes: i64,
+}
+
+impl ChangeSet {
+    fn from_entries(mut entries: Vec<ChangeEntry>) -> Self {
+        entries.sort_by(|left, right| left.path.cmp(&right.path));
+        let mut set = Self {
+            added: entries.iter().filter(|e| e.change == Change::Added).count(),
+            modified: entries
+                .iter()
+                .filter(|e| e.change == Change::Modified)
+                .count(),
+            deleted: entries
+                .iter()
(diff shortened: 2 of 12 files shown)


BunsDev:
Filed the commit bead: **`coven-fty`** — *Implement coven-afs commit materialization*.

Answering the question in the PR body — commit stays **deferred out of this PR** rather than folded in. The 501 + `afsCommit:false` capability handshake is the right shape to land now; commit materialization is a git-worktree-plus-signing, all-or-nothing operation and deserves its own change, as you argued.

Scope of `coven-fty`, from DESIGN.md §5:
- The eight materialization steps — quiesce, base verification, worktree creation, change-set application, escape refusal, signed commit, trailer attribution, `afs_commit` record
- The four error codes this PR deliberately left unconstructed (`afs.base_diverged`, `afs.path_outside_root`, `afs.copy_up_too_large`, `afs.commit_conflict`) plus `afs.commit_unsigned`, each arriving with the code path that raises it so clippy stays quiet
- Flipping `afsCommit` to `true` and removing the `afs.commit_unsupported` stub

Linked `discovered-from: coven-5kt` and `blocks: coven-gr1`, and noted on `coven-5kt` so its "NO BEAD FILED YET" item is resolved. `mount` remains deferred separately under `coven-x77`.

One unrelated note on CI: the two failed runs on this branch both failed only on `daemon::tests::native_detached_pty_timeout_persists_failed_exit_and_kills_tree` (Windows, 1688 passed / 1 failed). This PR doesn't touch `crates/coven-cli/src/daemon.rs`, and that test polls a 2s PTY timeout on a 10s deadline, so it reads as a loaded-runner flake rather than a regression. The current head (`09508a9`) has no CI result at all — the run is sitting in `action_required` because the last push came from the Copilot agent, so it needs a maintainer to approve the workflow.

PR #708, opened 2026-08-10
fix(test): stop the Windows PTY stub from disabling its own timeout

BunsDev:
Second attempt at `coven-5ua`. **The first one (#705) regressed into a worse
failure**, and reading the startup path explains both.

## What #705 got wrong

`spawn_detached_with_observer_and_timeout` races three claimants on a single
`AtomicU8` — meaningful output, child exit, and the timeout thread's
`compare_exchange(0 -> 2)`. **Only the winner acts.**
`MeaningfulOutputDetector` treats any printable ASCII byte in Ground state as
meaningful, so one stray byte reaching the PTY claims the state permanently:
the timeout never fires, the stub blocks forever in `expect_reply`, `on_exit`
never runs, and every watchdog in the test expires.

That is exactly the new failure — `Error: timed out waiting on channel` on CI
run 31370145402, against a diff (#706) touching only the workflow, a script,
and docs.

**Raising the startup timeout to 15s gave that stray byte 7.5× longer to win
the race.** Short is safer here, not riskier. I had it backwards.

## What this changes

- **Startup timeout back to 2s.** The pid-file race stays fixed by #705's
  polling, which is what actually addressed it — the pid is caught the instant
  it is written rather than read after the kill. The longer timeout was never
  the load-bearing part.
- **The stray byte loses its source.** The stub's descendant is
  `cmd.exe ... >nul` spawned with *inherited* stdio, so anything it emits lands
  on the PTY. This mode exists to produce no meaningful output, so its stdio now
  goes to null — the fixture now matches what it claims to model. The descendant
  is also spawned and registered before the terminal is touched, so the pid
  reaches disk as early as possible.
- **The failure diagnoses itself.** The startup timeout emits its own message
  when it wins, so its absence proves something else claimed the state. The
  `recv_timeout` path now reports which of the two defects occurred and the
  observed output, instead of only that a deadline passed.

## Verification

fmt clean, clippy clean workspace-wide with `-D warnings`, 51 `pty_runner`
tests pass, secret scan clean.

⚠️ **The changed code is `cfg(windows)` and cannot be type-checked on macOS** —
cross-checking against `x86_64-pc-windows-msvc` fails in the `libsqlite3-sys`
and `ring` build scripts, which need an MSVC-targeting C compiler. CI's Windows
job is the check.

**One green Windows run is not evidence this is fixed** — #705 had one and
regressed on the next PR. The bead stays open until it survives several runs.

### DIFF
--- a/crates/coven-cli/src/pty_runner.rs
+++ b/crates/coven-cli/src/pty_runner.rs
@@ -3746,19 +3746,24 @@ mod tests {
         Ok(())
     }
 
-    /// Headroom for the fixture to register its descendant, not a measured
-    /// budget.
+    /// Deliberately short, and shorter is safer here.
     ///
-    /// The stub spawns a `cmd.exe` descendant, writes its pid, and only then
-    /// blocks on a CPR reply that never arrives — and all of that has to
-    /// happen before the startup timeout under test fires. Two seconds was
-    /// enough on an idle host and not on a loaded CI runner, where the runner
-    /// killed the stub before the pid file existed and the test failed reading
-    /// it (`coven-047`, `coven-5ua`). The behaviour under test is that a
-    /// startup timeout fails the run and kills the descendant; how long the
-    /// timeout is set to is not part of that claim.
+    /// `spawn_detached_with_observer_and_timeout` races three claimants on one
+    /// `AtomicU8`: meaningful output, child exit, and the timeout thread's
+    /// `compare_exchange(0 -> 2)`. Only the winner acts, so if anything
+    /// printable reaches the PTY first the timeout never fires at all — the
+    /// stub then blocks forever in `expect_reply`, `on_exit` never runs, and
+    /// every watchdog in this test expires.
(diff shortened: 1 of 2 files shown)


BunsDev:
Verified on five Windows runs:\n\n- 32089387861 / 95568439764 — target test passed; full suite passed in 267.62s\n- 32085860093 / 95558128326 — target test passed; full suite passed in 211.37s\n- 32085446209 / 95556915424 — target test passed; full suite passed in 232.72s\n- 32084678374 / 95554642716 — target test passed; full suite passed in 184.05s\n- 32018125587 / 95351870731 — target test passed; full suite passed in 255.59s\n\nMerge commit 5f91a6be92d3a9574017355dc2665618771ff579 is on origin/main. The root-cause fix is present: descendant stdio no longer leaks into the PTY, startup timeout is back to 2s, and watchdog failures now report which race claimant won. Closing coven-5ua.

PR #722, opened 2026-08-10
fix(windows): hide noninteractive harness processes

### LINKED ISSUE #712: Windows noninteractive sessions open a console window
## What happened
On Windows, the detached Coven daemon launches a console-subsystem harness for `launchMode: nonInteractive` through `spawn_piped_with_observer` without `CREATE_NO_WINDOW`. A visible console window can therefore be allocated for a headless client session. Cave Research Desk with a local Codex familiar is a concrete reproducer.

The npm wrapper also launches the native Coven binary with `windowsHide: false`, so daemon cold-start can expose the same class of window before the daemon-backed harness starts.

## Steps to reproduce
1. Install `@opencoven/cli` on Windows and authenticate Codex.
2. Start the daemon without an owning console, or let Coven Cave start it.
3. POST a Codex session with `launchMode: nonInteractive`.
4. Observe a console window allocated for the console-subsystem child.

## Expected behavior
Noninteractive sessions launched by a detached daemon run without allocating a visible console. Interactive PTY sessions retain their existing terminal behavior. Wrapper-to-native launch, cancellation, timeout, signals, and output propagation remain intact.

## Version and harness
- Reproduced against npm `@opencoven/cli@0.2.5`; the missing creation flag remains on current `main` at initial audit.
- Harness: Codex; the affected spawn boundary is shared by Windows noninteractive harnesses.

## Implementation readiness
Centralize Windows noninteractive creation flags, cover the `CREATE_SUSPENDED` process-tree path without weakening job containment, and validate the complete wrapper/daemon/harness descendant tree with a Windows-native console-visibility fixture. Do not change interactive PTY semantics.

### LINKED ISSUE #713: Noninteractive sessions need an explicit unattended write contract
## What happened
A daemon-backed Research Desk Codex session completed every reasoning phase and exited 0, but produced no required artifact. The session ran with `approval: untrusted`; every file and shell write was denied because command/file approval is unsupported in one-shot exec mode.

Cave currently sends only `projectRoot`, `harness`, `prompt`, and `launchMode`. `SessionLaunch` has no explicit unattended approval or workspace-write contract, so a noninteractive client cannot request the narrow permissions needed to create its declared outputs.

## Steps to reproduce
1. Configure Codex with `approval_policy = "untrusted"`.
2. POST a local `launchMode: nonInteractive` session whose prompt requires writing a file under `projectRoot`.
3. Observe Codex decline each write because exec mode cannot ask for approval.
4. The process exits 0 without the requested artifact.

## Expected behavior
A trusted local client can explicitly request an unattended, workspace-confined launch policy. The daemon validates the request, maps it to approval-never plus a writable workspace sandbox, and never widens access beyond the canonical project root and validated additional directories. Untrusted/webhook clients cannot silently opt into it.

## Security and compatibility
This must be an explicit API contract, not an environment or prompt convention. The Rust daemon remains the authority boundary. Existing clients and launches keep their current defaults. Unsupported daemon versions fail closed. Add API, harness-argv, boundary, and real artifact-write regressions.

CompleteDotTech:
## Context

- Summary: Repair the Windows Research Desk process boundary end to end: hide daemon-backed noninteractive harnesses, contain and reap their full process trees, make prompt delivery bounded and cancellable, and expose an exact native supervisor contract for desktop clients.
- Root cause: Cave PR #4413 repaired Node-side child spawning, but Research Desk's local Codex familiar enters Coven with `launchMode: "nonInteractive"`. Coven's Rust `spawn_piped_with_observer` path did not apply `CREATE_NO_WINDOW`; the npm wrapper also used inherited stdio, for which Node/libuv cannot combine `windowsHide` with `CREATE_NO_WINDOW`. Process registration, prompt delivery, and post-spawn Job assignment left additional cancellation and shutdown races.
- Native follow-up root causes: Codex 0.146.1 downgrades requested `workspace-write` to `read-only` on native Windows when `windows.sandbox` is disabled. It also parses root `--ask-for-approval never` without inheriting that field into `exec`; when AutoReview becomes effective, `exec` rebuilds without its synthetic headless default and an untrusted project can select approval `untrusted`. Coven now pins both `-c windows.sandbox="unelevated"` and the session-level `-c approval_policy="never"` inside its validated Windows Codex `nonInteractive` `launchPolicy` boundary.
- Files changed: Rust daemon/API/harness/process-runner code and native fixtures; the npm wrapper and its tests; daemon/API/supervisor documentation and release notes.
- Closes #712
- Closes #713

## Implementation

- Centralizes strict noninteractive containment. Windows children start suspended and hidden, enter `KILL_ON_JOB_CLOSE` containment before resume, and inherit a daemon-lifetime Job from process creation; Unix children keep their process-group behavior and now shut down gracefully through bounded runtime cleanup.
- Makes explicit kill a quiescence boundary: descendants are terminated and reaped, stdout/stderr drains and exit callbacks complete in order, failed kills retain exact ownership for retry, and concurrent cancel cannot be overwritten by a later launch failure.
- Starts output drains before bounded prompt delivery, registers exact ownership before writing, moves full Codex prompts to stdin, and fails deterministically when delivery or admission fails. Large duplex, never-read, closed-stdin, root-exit, cancellation, timeout, and daemon-stop races have executable regressions.
- Adds `RequestAuthority::{OwnerLocalIpc,Tcp}`. Existing public/internal handlers remain owner-local by default; the daemon HTTP transport supplies authority explicitly. `launchPolicy` is available only over owner-gated Unix socket/Windows pipe transport, TCP health advertises `sessionLaunchPolicy: false`, and TCP POSTs containing it are rejected before session creation or spawn. Canonical existing `addDirs` outside `projectRoot` remain valid grants and are forwarded exactly.
- Resolves official Windows Codex npm shims to the validated architecture-specific native executable for noninteractive launches, preserving managed-package markers and exact Unicode/metacharacter argv without `cmd.exe`. Interactive PTY/stream modes retain the shim; a real `codex.cmd` ConPTY fixture proves terminal query/reply behavior is unchanged.
- Selects Codex's supported Windows unelevated restricted-token sandbox backend only for the exact daemon-authorized noninteractive workspace-write policy. The adapter also pins `approval_policy="never"` as an explicit Codex session config override so AutoReview cannot replace the requested policy during `exec` configuration. Interactive/user-owned sessions, stream sessions, non-Codex harnesses, and non-Windows platforms receive no backend override. Backend failures retain Codex stderr/exit evidence and never trigger a shell or unsandboxed fallback.
- Adds hidden `coven process-supervisor --protocol coven.process-supervisor.v1` plus wrapper-only `--print-native-binary-path`. The JSONL admission frame, stdin ownership lease, ready/error record, output forwarding, exit propagation, parent-death behavior, frame cap, and absolute-path checks are documented and tested. No shell, string evaluation, or PID/taskkill fallback is introduced.
- Changes the wrapper no-window opt-in to pipe-backed transparent forwarding so libuv can apply `CREATE_NO_WINDOW`; consumes the opt-in environment key before native spawn and leaves ordinary interactive wrapper behavior unchanged.

### User-visible behavior

- Detached Windows Research/Codex sessions no longer allocate a visible console window.
- The Windows Codex Research policy now remains effectively `workspace-write` instead of being silently reduced to `read-only` by a disabled native sandbox backend.
- Cancellation, timeout, app/daemon shutdown, and direct desktop-client process ownership stop the complete descendant tree before reporting completion.
- Oversized Research prompts no longer ride in one Windows command line, and missing primary artifacts remain a distinct fail-closed application condition.

### Compatibility notes

- Interactive PTY behavior is unchanged and has a native `codex.cmd`/ConPTY regression.
- macOS/Linux do not receive Windows creation flags or the Windows Codex backend override. Both Apple targets compile; native macOS runtime access was unavailable and is not claimed.
- Windows daemon startup deliberately fails closed if its daemon-lifetime Job cannot be created, configured, or assigned. Allowing degraded startup would reopen the pre-session Job-assignment escape window this repair closes.
- The supervisor is an internal desktop-integration contract; ordinary CLI invocation remains unchanged.

## Verification

- [x] `cargo fmt --check`
- [x] Focused `launch_policy` tests: 9 passed; process-supervisor integration: 7 passed
- [x] `cargo test --workspace --locked` (Linux: 1,940 CLI unit tests passed, 2 ignored; every workspace/integration suite passed)
- [x] Combined Node wrapper/release/benchmark guardrails: 173 passed, 1 Windows-only skip on Linux
- [x] `COVEN_NPM_DRY_RUN_VERSION=999.0.0 node scripts/test-cli-prepublish.mjs --target=linux-x64 --skip-secrets-scan` (all four package build/pack/install/daemon-lifecycle checks passed; dry-run only)
- [x] `python3 scripts/check-api-contract-docs-test.py` (48 passed) and `python3 scripts/check-api-contract-docs.py`
- [x] `python3 scripts/check-secrets-test.py` (87 passed) and `python3 scripts/check-secrets.py`
- [x] `python3 scripts/check-coven-privacy-test.py` (33 passed) and `python3 scripts/check-coven-privacy.py --range origin/main...HEAD`
- [x] `git diff --check origin/main...HEAD`
- [x] Apple compile-only: `cargo check -p coven-cli --locked --target x86_64-apple-darwin` and `--target aarch64-apple-darwin` via Zig cross linker

### Native Windows

- Exact head `4cd2de46c801c74b50d32a84fdbaeaca4de523fb` rebuilt cleanly on native Windows: `coven.exe` is 32,505,856 bytes with SHA-256 `7216fa34e5b30531c6eb51ce92915a6d87070358251543b1bc766d5a39171e16`. Two identical owner-local `launchPolicy` sessions with AutoReview enabled both persisted effective approval `never` plus sandbox `workspace-write`, completed with exit 0, and wrote the exact 25-byte sentinel; the second-session regression that previously became `untrusted` is closed.
- `cargo test -p coven-cli --locked --bin coven windows -- --nocapture --test-threads=1`: 34 passed, including the allocated-console positive control, hidden piped child, direct official Codex native argv/no-console path, strict Job flags, Job-close descendant backstop, and real interactive `codex.cmd` ConPTY flow.
- `cargo test -p coven-cli --locked --test stream_json_integration windows_ -- --nocapture --test-threads=1`: 2 passed (official shim-to-native stream/resume and silent timeout durability).
- `cargo test -p coven-cli --locked --test process_supervisor --test windows_daemon_lifecycle -- --nocapture --test-threads=1`: supervisor 5/5; captured native/wrapper daemon start and abrupt shutdown 2/2.
- Windows Node `--test scripts/publish-npm-test.mjs`: 69 passed, 0 failed, 3 platform-inapplicable skips; the opt-in wrapper positive-control test covered hidden console, stdin/stdout/stderr forwarding, exit, broken destination, and Ctrl-C behavior.
- Coordinated Cave provider contract suite against the native binary: 30/30, including owner EOF, descendant quiescence, and abrupt owner death.
- Native Codex 0.146.1 discriminator, isolated to fresh session/rollout IDs and the intended cwd: both policy flag orderings without a selected Windows backend reported approval `never`, sandbox `read-only`, and wrote no sentinel; adding only `-c windows.sandbox="unelevated"` reported sandbox `workspace-write` and wrote the exact sentinel. This is direct platform evidence for the targeted override now emitted by Coven.

### CI

- Current exact head: `4cd2de46c801c74b50d32a84fdbaeaca4de523fb`.
- Exact-head CI: [run 31449780875](https://github.com/OpenCoven/coven/actions/runs/31449780875) completed successfully; all 13 CI jobs passed on the exact signed head.
- Ancillary policy workflow: [Dependabot auto-merge run 31437332441](https://github.com/OpenCoven/coven/actions/runs/31437332441) (expected skipped for a non-Dependabot PR).

## Risk and Rollback

- Risk level: high, because this changes process ownership and shutdown ordering across daemon, wrapper, and desktop-client boundaries. The implementation keeps platform policy centralized and adds native fixtures at every affected boundary.
- Rollback plan: revert the single signed commit. Existing clients default to the unchanged owner-local handler contract; no storage migration or release artifact is involved.
- Observed performance boundary: the native unelevated Codex sandbox was functionally correct, but simple PowerShell tool calls took minutes during one real Research run. That latency is recorded as a separate native/runtime limitation; this PR does not claim to repair it.

## Agent Handoff

- Current state: one SSH-signed, DCO-compliant commit based directly on current `origin/main`; an independent provider audit of the Windows backend repair reported no blockers before publication.
- Cross-repository dependencies: coordinated Coven Cave lifecycle PR [OpenCoven/coven-cave#4524](https://github.com/OpenCoven/coven-cave/pull/4524); companion Cave launch-hardening PR [OpenCoven/coven-cave#4531](https://github.com/OpenCoven/coven-cave/pull/4531); approval-policy verification follow-up [OpenCoven/coven-cave#4537](https://github.com/OpenCoven/coven-cave/pull/4537), squash-merged externally at verified Cave `main` `429a28dd` with a tree equal to audited head `040a28c`.
- Known gaps: no native macOS runtime machine was available. Apple compile proof is recorded above; Windows native process/console and effective sandbox behavior are covered directly. The PowerShell latency boundary above remains explicit.
- No merge, package publication, tag, or release occurred as part of this work.

### DIFF
--- a/crates/coven-cli/src/api.rs
+++ b/crates/coven-cli/src/api.rs
@@ -21,8 +21,8 @@ use crate::{
     daemon::DaemonStatus,
     encrypted_artifacts::SensitiveArtifactStore,
     handoff::{HandoffPacketV1, WorkspaceSnapshot},
-    harness::{ConversationHint, HarnessLaunchMode},
-    privacy, session_launch, store, ward,
+    harness::{ConversationHint, HarnessLaunchMode, LaunchPolicy},
+    privacy, project, session_launch, store, ward,
 };
 
 const MAX_EVENTS_LIMIT: i64 = 1_000;
@@ -127,6 +127,10 @@ pub struct HealthCapabilities {
     pub event_cursor: String,
     pub structured_errors: bool,
     pub session_handoff: bool,
+    /// Whether `POST /sessions` accepts the exact, fail-closed
+    /// `launchPolicy` contract documented for unattended Codex work.
+    #[serde(default)]
+    pub session_launch_policy: bool,
     /// Whether the `afs.*` route family is served at all.
     pub afs: bool,
     /// Mount backend, or `false` when none is available. A client must branch
@@ -239,6 +243,7 @@ pub struct SessionLaunch {
     /// transform for the underlying CLI.
     pub model: Option<String>,
     pub launch_mode: HarnessLaunchMode,
+    pub launch_policy: Option<LaunchPolicy>,
     pub prompt: String,
     pub title: String,
     pub conversation: Option<ConversationHint>,
@@ -333,7 +338,29 @@ impl SessionRuntime for NoopSessionRuntime {
     }
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum RequestAuthority {
+    /// Filesystem-permission-protected Unix socket or owner-only Windows pipe.
+    OwnerLocalIpc,
+    /// Optional loopback TCP listener. Host/Origin checks reduce browser risk,
+    /// but they do not prove that the caller owns the daemon process.
+    Tcp,
+}
+
+impl RequestAuthority {
+    fn allows_session_launch_policy(self) -> bool {
+        matches!(self, Self::OwnerLocalIpc)
+    }
+}
+
 pub fn health_response(daemon: Option<DaemonStatus>) -> HealthResponse {
+    health_response_for_authority(daemon, RequestAuthority::OwnerLocalIpc)
+}
+
+pub(crate) fn health_response_for_authority(
+    daemon: Option<DaemonStatus>,
+    authority: RequestAuthority,
+) -> HealthResponse {
     HealthResponse {
         ok: true,
         api_version: COVEN_API_NAMED_VERSION.to_string(),
@@ -348,6 +375,7 @@ pub fn health_response(daemon: Option<DaemonStatus>) -> HealthResponse {
             event_cursor: "sequence".to_string(),
             structured_errors: true,
             session_handoff: true,
+            session_launch_policy: authority.allows_session_launch_policy(),
             afs: true,
             afs_mount: MountCapability::detect(),
             afs_commit: true,
@@ -364,8 +392,9 @@ fn health_response_with_hub(
     coven_home: &Path,
     daemon: Option<DaemonStatus>,
     event_writer: Option<crate::event_writer::EventWriterHealth>,
+    authority: RequestAuthority,
 ) -> HealthResponse {
-    let mut response = health_response(daemon);
+    let mut response = health_response_for_authority(daemon, authority);
     if let Ok(summary) = crate::hub::hub_health_summary(coven_home) {
         response.hub = serde_json::from_value(summary).ok();
     }
@@ -413,6 +442,26 @@ pub fn handle_request_with_runtime(
     body: Option<&str>,
     runtime: &dyn SessionRuntime,
 ) -> Result<ApiResponse> {
+    handle_request_with_runtime_and_authority(
+        method,
+        path,
+        coven_home,
+        daemon,
+        body,
+        runtime,
+        RequestAuthority::OwnerLocalIpc,
+    )
+}
+
+pub(crate) fn handle_request_with_runtime_and_authority(
+    method: &str,
+    path: &str,
+    coven_home: &Path,
+    daemon: Option<DaemonStatus>,
+    body: Option<&str>,
+    runtime: &dyn SessionRuntime,
+    authority: RequestAuthority,
+) -> Result<ApiResponse> {
     let (route, query) = split_path_query(path);
     let route = match normalize_api_route(route) {
         ApiRoute::Route(route) => route,
@@ -441,7 +490,7 @@ pub fn handle_request_with_runtime(
         ),
         ("GET", "/health") => json_response(
             200,
-            &health_response_with_hub(coven_home, daemon, runtime.event_writer_health()),
+            &health_response_with_hub(coven_home, daemon, runtime.event_writer_health(), authority),
         ),
         ("GET", "/capabilities") => json_response(200, &control_plane::capabilities()),
         ("POST", "/afs/sessions") => afs_create(coven_home, body),
@@ -791,7 +840,7 @@ pub fn handle_request_with_runtime(
             get_scheduler_loop_state(coven_home, loop_id)
         }
         ("GET", "/sessions") => list_sessions_response(coven_home, query),
-        ("POST", "/sessions") => launch_session(coven_home, body, runtime),
+        ("POST", "/sessions") => launch_session(coven_home, body, runtime, authority),
         ("POST", "/sessions/external") => register_external_session(coven_home, body),
         ("POST", path) if path.starts_with("/sessions/") && path.ends_with("/complete") => {
             let session_id = session_action_id(path, "/complete");
@@ -1856,6 +1905,7 @@ fn launch_session(
     coven_home: &Path,
     body: Option<&str>,
     runtime: &dyn SessionRuntime,
+    authority: RequestAuthority,
 ) -> Result<ApiResponse> {
     // Client-side validation errors (malformed JSON, bad fields,
     // unsupported launchMode, malformed `conversation` object, …) must
@@ -1868,6 +1918,14 @@ fn launch_session(
             return api_error(400, "invalid_request", &error.to_string(), None);
         }
     };
+    if payload.get("launchPolicy").is_some() && !authority.allows_session_launch_policy() {
+        return api_error(
+            403,
+            "forbidden",
+            "launchPolicy is accepted only over the owner-gated local IPC transport.",
+            None,
+        );
+    }
     let mut launch = match session_launch_from_payload(payload) {
         Ok(launch) => launch,
         Err(error) => {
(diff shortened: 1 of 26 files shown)


BunsDev:
Reviewed the Rust side — mainly the Windows containment path, the new
capability gating, and the teardown ordering. This is careful work and the
tests are unusually good for a change this size. Two things I'd raise, one of
which I think is worth changing before merge.

## The daemon now refuses to start if the lifetime Job cannot be assigned

`serve_forever` calls `install_daemon_lifetime_job()?`, so a failed
`AssignProcessToJobObject` aborts daemon startup entirely.

The doc comment on that function describes the lifetime Job as birth-time
containment "for the narrow interval before a suspended noninteractive child is
assigned to its per-session Job", and says the per-session Job "remains the
checked, explicit cancellation owner during normal operation." If that's right,
this is a backstop — and a backstop being unavailable shouldn't cost the whole
daemon.

The failure is reachable: a daemon already inside a Job that doesn't permit
nesting can't be assigned, which is why the error text says "the parent Job must
permit modern nested jobs." That covers CI runners, some MDM/task-scheduler
launches, and container-ish setups. In those environments `coven daemon serve`
now fails outright, and the operator's symptom is "the daemon won't start" with
a Win32 error, rather than "noninteractive containment is degraded."

Suggestion: warn and continue rather than propagate, so the per-session Job
still does its job and the failure is visible without being fatal. If failing
closed *is* deliberate — i.e. you'd rather not run at all than run without
birth-time containment — that's a defensible call, but it deserves a sentence
saying so, because the current comment argues the opposite.

## That path has no test

The success path is well covered. I couldn't find a test exercising assignment
failure, so whichever behaviour you choose (fatal or degraded) isn't pinned.
Worth one, since it's the branch that only fires on machines you don't develop
on.

## Things I checked and liked

- `terminate_tree` marks `terminated` even when `TerminateJobObject` reports an
  error, on the grounds that closing the `KILL_ON_JOB_CLOSE` handle is the
  kernel-enforced backstop. That's the right call and the comment explains why,
  which is what stops someone "fixing" it into a retry loop later.
- `sessionLaunchPolicy` is gated to owner-local IPC with TCP rejecting the field
  *and* the capability absent from TCP health — both directions tested. Clients
  can't discover it where they can't use it.
- Adding `check-api-contract-docs.py` alongside the contract change is the part
  I'd have most wanted and least expected.

Nothing here blocks on my account; I can't run Windows locally, so treat the
first point as a question about intent rather than a verified defect.

PR #783, opened 2026-08-22
feat(client): reach the daemon's session page cursor

CompleteDotTech:
## Audit first: what the session/event read surface already is

Cave's Chat v1 Phase 2 (`cave-g9d49`, OpenCoven/coven-cave#4835) needs three
canonical reads: list sessions, read one session, and read that session's
events with stable pagination. This started as an audit of whether #760 left
any of them missing. It did not leave much:

**Daemon (`crates/coven-cli/src/api.rs`) — complete.**

| Read | Route | Handler |
|---|---|---|
| List sessions | `GET /api/v1/sessions` | `list_sessions_response`, `api.rs:4291` |
| One session | `GET /api/v1/sessions/:id` | route arm, `api.rs:962` |
| Events | `GET /api/v1/events?sessionId=…` and `GET /api/v1/sessions/:id/events` | `list_session_events`, `api.rs:4557` |

Session paging is real (`store::list_session_page`, `store.rs:2728`): keyset
ordering on `created_at, id` with an opaque URL-safe base64 cursor
(`store.rs:2790`), a 1–1000 `limit`, and `includeArchived`. Event paging is
real too, with a byte-bounded page builder and `nextCursor.afterSeq` /
`hasMore` (`api.rs:4633`).

**Typed client (`crates/coven-client`) — one gap.** `ReadEndpoint` covered
`Session`, `Sessions { limit }`, and `Events { after_seq, limit }`
(`models.rs:29`). Sessions could not express `cursor` or `includeArchived`, so
no `coven-client` caller could ask for a second page. The one in-tree consumer
requested 100 and threw `nextCursor` away
(`tui/chat/client.rs:486`), so a ledger past 100 sessions silently lost the
remainder from the chat overlay.

The v0.4.1 release program (#780) does not schedule this; its `coven-v041-cave`
task is daemon socket/timeout hardening, unrelated.

## What this changes

- `ReadEndpoint::Sessions` carries `cursor` and `include_archived` beside
  `limit`. Setting none of the three still requests the inherited unpaginated
  array, so existing callers are unchanged.
- No capability flag gates the cursor, and deliberately so: a caller can only
  send a cursor the daemon itself just issued, so it is self-gating in a way
  `afterSeq` is not. The client does validate it against the daemon's own
  base64 alphabet before it reaches the request target, so a corrupted or
  hand-composed value cannot smuggle a separator into the inherited raw v1
  query.
- `collect_session_pages` follows `nextCursor` to exhaustion, mirroring the
  existing `consume_event_pages`. It refuses a repeated cursor and a
  continuation on an empty page rather than replaying one forever, and stops at
  a stated ceiling instead of an unstated 100.
- `docs/API-CONTRACT.md` documented this route as `SessionRecord[]` and nothing
  else — the envelope, the query, and the cursor were entirely undocumented. Now
  both shapes are described.

## Known and deliberately out of scope

`afterEventId` is served by the daemon (`api.rs:4572`) and documented as a
compatibility cursor, but `ReadEndpoint::Events` still cannot express it.
`afterSeq` covers Phase 2, and adding the field would churn every `Events`
call site for no current caller.

## Validation

| Check | Result |
|---|---|
| `cargo fmt --all -- --check` | clean |
| `cargo clippy -p coven-client -p coven-cli --all-targets` | no new findings; nothing on the changed files |
| `cargo test -p coven-client --lib` (Windows) | 64 passed, incl. 2 new path/cursor tests |
| `cargo test -p coven-client --test health` (Linux) | 40 passed, incl. 2 new integration tests |
| `cargo test -p coven-cli --bin coven -- tui::chat::client` | 10 passed, incl. 3 new paging tests |
| `cargo test -p coven-cli --bin coven -- api::tests::` | 363 passed |
| `cargo test -p coven-cli --bin coven` (full) | 2098 passed / 16 failed |

The 16 failures are pre-existing Windows-environment failures, not regressions:
the same modules fail identically on unmodified `origin/main` (`afs::tests`
symlink creation needs a privilege this host does not hold, `daemon::tests`
named-pipe ACL, `maintenance_gate::tests`, `pty_runner::tests`), and the count
varies run to run on the PTY cases. None touch the changed files. The
`coven-client` unix tests are `#![cfg(unix)]` and were compiled and run under
WSL, which is what caught three call sites Windows could not see.

Refs OpenCoven/coven-cave#4835

### DIFF
--- a/crates/coven-cli/src/tui/chat/client.rs
+++ b/crates/coven-cli/src/tui/chat/client.rs
@@ -98,6 +98,14 @@ pub(crate) struct ChatEventQuery<'a> {
 pub(crate) const EVENT_PAGE_LIMIT: i64 = 512;
 pub(crate) const EVENT_PAGES_PER_POLL: usize = 8;
 
+/// Sessions requested per `/sessions` round trip. Stays inside the daemon's
+/// 1000-row page ceiling so a page is one bounded response, not a guess.
+pub(crate) const SESSION_PAGE_LIMIT: u16 = 200;
+/// Ceiling on the overlay listing. The overlay renders from memory, so the
+/// listing stops paging here rather than following an unbounded ledger; the
+/// previous single request stopped at 100 without saying so.
+pub(crate) const MAX_LISTED_SESSIONS: usize = 2_000;
+
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub(crate) struct ChatEventPage {
     pub(crate) events: Vec<store::EventRecord>,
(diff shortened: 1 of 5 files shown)


CompleteDotTech:
CI note, for the record: on the first run of head \`7da0d45\`, `Rust tests (Windows)` failed once on `api::tests::adopted_launch_concurrent_replay_observes_committed_created` — a request-adoption concurrency test that touches nothing in this diff, and that passed on the previous head carrying byte-identical Rust (only the docs sample changed between the two). A rerun of the failed jobs went green. All checks now pass on `7da0d45`.

PR #789, opened 2026-08-22
docs: define the mobile device trust architecture

BunsDev:
## Summary

- defines the canonical owner / installation / familiar / device / grant / session separation
- records the existing `mobile_memory` v1 implementation as the migration foundation rather than replacing it
- specifies pairing transcript binding, device grants, biometric step-up, relay constraints, recovery, attestation, and threat-model invariants
- maps #785–#788 into an additive implementation sequence with explicit compatibility and merge gates

## Readiness packet

**Scope:** architecture and migration contract only; no runtime behavior changes.

**Current implementation audited:**
- `crates/coven-cli/src/mobile_memory/pairing.rs`
- `crates/coven-cli/src/mobile_memory/auth.rs`
- `crates/coven-cli/src/mobile_memory/registry.rs`
- `crates/coven-cli/src/mobile_memory/gateway.rs`
- `crates/coven-relay`

**Verification:** documentation-only change; repository CI is the merge gate.

Closes #784

Issue #787, opened 2026-08-22
Plan: Seamless trusted-device reconnection, local discovery, and relay fallback


BunsDev:
Parent architecture: #784
Builds on: #785, #786

## Goal

Make the QR a first-introduction ceremony rather than a recurring login step. Once enrolled, trusted phones should reconnect to Coven/Psyche with minimal friction while preserving cryptographic authentication, privacy, and explicit authorization.

Target returning-device UX:

```text
$ coven

Nearby trusted device: Val’s iPhone
Waiting for approval…
✓ Authenticated
```

The actual proof remains device-key challenge/response. Biometric/user verification is invoked according to the assurance policy defined in #786.

### Phase 1 — relay-first

- [ ] Maintain/re-establish outbound encrypted connectivity through the rendezvous/relay path.
- [ ] Authenticate returning devices using fresh challenges and proof of possession.
- [ ] Rotate ephemeral session keys on every connection.
- [ ] Ensure the relay cannot impersonate either endpoint or decrypt application traffic.

### Phase 2 — local fast path

- [ ] Discover compatible Coven/Psyche nodes on the LAN using mDNS/DNS-SD or an equivalent local mechanism.
- [ ] Advertise only opaque, rotating identifiers rather than owner/familiar names or stable device identifiers.
- [ ] Attempt authenticated direct LAN transport first when appropriate.
- [ ] Fall back automatically to the relay when direct connectivity is unavailable.
- [ ] Prevent local-network discovery from becoming authentication; discovery only finds a candidate endpoint.

### Phase 3 — proximity-assisted introduction

Research BLE as optional proximity evidence for first-contact/nearby enrollment. Do not carry sensitive application traffic over BLE merely to avoid the established encrypted transport.

Potential UX:

```text
OpenCoven Mobile

Nearby Cave
“MacBook Pro”

[Connect with Face ID]
```

For first enrollment, preserve a confirmation/authentication ceremony unless another existing trusted relationship establishes the introduction.

## Session lifecycle

- [ ] Define reconnect/resumption tokens that are sender-constrained or cryptographically bound rather than reusable bearer secrets.
- [ ] Define offline, suspended, expired, revoked, and key-rotated states.
- [ ] Ensure resumed sessions cannot bypass newly tightened grant scopes.
- [ ] Handle relay migration/network changes without changing device identity.
- [ ] Ensure one device can safely connect to multiple Covens without cross-Coven correlation.

## Push-based approval

[...]

Issue #785, opened 2026-08-22
Plan: TUI QR bootstrap and end-to-end encrypted mobile pairing


BunsDev:
Parent architecture: #784

## Goal

Make first-time mobile enrollment from the Coven/Psyche TUI fast, local-first, and cryptographically authenticated. The QR is an **out-of-band introduction**, not a reusable login token.

Target UX:

```text
$ coven device pair

Pair OpenCoven Mobile
Requesting:
  ✓ View sessions
  ✓ Send messages
  ✕ Execute tools without approval
  ✕ Export identity or memory

[ QR CODE ]

Expires in 01:47
```

## Pairing offer

Define a compact versioned offer containing approximately:

```text
PairingOffer {
  version
  pairing_session       // cryptographically random, single-use
  ephemeral_public_key
  rendezvous_hint
  local_discovery_hint?
  requested_capabilities_hash
  expires_at
}
```

Prefer canonical CBOR plus compact URL-safe encoding over verbose JSON. Support an OpenCoven Universal Link/App Link so the camera can hand the offer directly to the installed mobile client.

The QR MUST NOT contain permanent API tokens, reusable bearer credentials, root/familiar private keys, biometric data, or unnecessary identity metadata.

## Handshake

- [ ] Generate a fresh ephemeral keypair for every pairing attempt.
- [ ] Establish an authenticated, forward-secret E2EE handshake using an established protocol/construction (evaluate Noise patterns rather than inventing cryptography).
- [ ] Bind the QR offer, ephemeral keys, requested capabilities, both endpoint identities, nonce, and protocol version into the handshake/enrollment transcript.
- [ ] Require explicit confirmation before issuing a durable device grant.
- [ ] Destroy pairing secrets after success, expiry, cancellation, or bounded failed attempts.
- [ ] Make pairing sessions short-lived and single-use.
- [ ] Add replay, substitution, MITM, downgrade, relay, and malformed-input tests.

## Rendezvous/transport MVP

Use outbound connections from both endpoints so pairing works across NAT, SSH hosts, and restrictive networks:

```text
TUI ── outbound ──▶ rendezvous/relay ◀── outbound ── Mobile
         ╲____________ E2EE _____________╱
```

The rendezvous service may match opaque pairing-session IDs and forward ciphertext but must not receive application plaintext or durable device secrets.

## Human verification

[...]

PR #790, opened 2026-08-22
feat(relay): add admission-controlled bounded mobile rendezvous

BunsDev:
## Summary

Adds the relay-first transport foundation for OpenCoven mobile connectivity:

- requires an operator-managed admission credential before a host can create a room
- gives host and client distinct canonical 256-bit room credentials
- prevents clients from creating rooms or occupying the host role
- forwards only opaque text/binary data; endpoint authentication and E2EE remain outside the relay
- bounds active rooms, message size, queue depth, per-peer queued bytes, and global queued bytes
- holds byte-budget permits through the actual WebSocket write
- delivers teardown through a channel independent of the bounded data queue
- applies idle and per-write deadlines so stalled peers cannot pin resources indefinitely
- zeroes in-memory room credentials when a room is removed
- documents WSS, edge rate limiting, credential handling, and relay metadata exposure

## Security boundary

This PR does **not** make relay credentials into OpenCoven identities or device grants. The relay is an opaque rendezvous transport. Hosts and clients must still establish the endpoint-authenticated encrypted pairing/session channel specified by the mobile-device trust architecture merged in #789.

The deployment admission credential is host-only and must never be included in the QR offer. A mobile client receives only its client room credential plus the independent endpoint-authentication material required by the inner protocol.

## Verification

The focused relay tests cover:

- canonical credential parsing
- host-only admitted room creation
- host/client role separation
- duplicate role rejection
- opposite-role forwarding
- per-peer and global queued-byte enforcement
- close delivery while the data queue is full
- stale unregister protection
- room cleanup and capacity bounds

Repository CI remains the merge gate:

```bash
cargo fmt --check
cargo clippy --workspace --all-targets -- -D warnings
cargo test --workspace --locked
python scripts/check-secrets.py
python3 scripts/check-coven-privacy.py --staged
```

## Follow-up scope

This is one implementation increment of #787, not the completion of the entire issue. Remaining work includes endpoint-authenticated pairing over the relay (#785/#786), seamless reconnect/session resumption, revocation propagation, privacy-preserving LAN discovery/direct fallback, and push-assisted wake-up/approval.

Refs #787

### DIFF
--- a/crates/coven-relay/README.md
+++ b/crates/coven-relay/README.md
@@ -1,9 +1,45 @@
 # coven-relay
 
-Stateless WebSocket relay for Hexes — bridges the Coven gateway running on a Mac to iOS Hexes clients.
+`coven-relay` is a bounded, ephemeral WebSocket rendezvous service for OpenCoven devices. It allows a Coven/Psyche host and a mobile client to meet through outbound connections when direct networking is unavailable.
 
-See `docs/specs/2026-05-31-hexes-implementation-plan.md` Track 2 for the full
-implementation roadmap.
+The relay is **not** an OpenCoven identity or authorization authority. It forwards opaque binary frames only. Endpoint authentication, device grants, transcript verification, and application encryption remain end to end between the host and client.
+
+## Protocol v1
+
+Connect to:
+
+```text
+/ws?v=1&room=<ROOM_ID>&role=<host|client>
+Authorization: Bearer <ROOM_CREDENTIAL>
+```
+
+`ROOM_ID` and `ROOM_CREDENTIAL` are separate canonical base64url encodings of 32 random bytes. They are short-lived rendezvous material, not durable device credentials.
+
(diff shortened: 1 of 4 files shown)


BunsDev:
Security review checkpoint:

- The policy core performs no cryptographic primitive construction and accepts canonical signed bytes plus verifier adapters.
- Unknown fields, capabilities, restrictions, versions, and state transitions fail closed.
- Device grants require both issuer-signature verification and proof of possession.
- Current revocation epoch and narrowed operation requirements are evaluated on every authorization.
- Exact transaction request/presentation digests, short TTLs, signatures, and single-use nonces are required.
- Signature verification occurs before nonce consumption to avoid attacker-driven nonce denial.
- The included nonce store is explicitly test-only; production adapters must be durable and transactional.
- Pairwise identifiers require a distinct owner-held binding secret and vary by trust domain.

This PR does not close the feature issues; it establishes the executable authorization contract for subsequent endpoint and transport PRs.

Issue #786, opened 2026-08-22
Plan: Device-bound credentials, biometric authorization, and scoped grants


BunsDev:
Parent architecture: #784
Depends on pairing bootstrap: #785

## Goal

After the one-time introduction, enroll the phone as a cryptographic device whose private key is protected by the platform and whose authority is expressed through a revocable OpenCoven grant.

**Biometrics remain local user verification. They are never transmitted or treated as the network credential.**

```text
Face ID / Touch ID / biometric
        ↓
OS verifies local user presence
        ↓
Secure hardware permits key operation
        ↓
Device signs fresh challenge/action digest
        ↓
Coven verifies signature + device grant
```

## Device credential

- [ ] Generate a non-exportable signing key on enrollment.
- [ ] iOS: evaluate Secure Enclave + Keychain access-control/LocalAuthentication policies.
- [ ] Android: evaluate Android Keystore hardware-backed keys + BiometricPrompt/strong-biometric policy.
- [ ] Treat platform PIN/passcode fallback separately from biometric-only policy; do not silently conflate assurance levels.
- [ ] Bind the device public key to the enrollment transcript from #785.
- [ ] Use a pairwise/pseudonymous device identity per Coven relationship where practical.

## Device grant

Specify and implement a signed grant approximately equivalent to:

```text
DeviceGrant {
  subject: hash(device_public_key)
  issuer: owner_or_coven_identity
  audience: coven_instance_or_trust_domain
  scopes: [...]
  restrictions: {...}
  issued_at
  expires_at
  revocation_epoch
}
```

Initial capability taxonomy should distinguish at minimum:

- session metadata/read
- conversation read
- message send
- tool invocation request
- tool execution approval
- secrets access
- familiar-memory access
- identity administration
- device enrollment/revocation
- identity/memory export

## Assurance levels

Avoid requiring biometrics for every packet. Define policy levels such as:

1. **Possession** — valid device key + grant; sufficient to resume low-risk connectivity.
2. **Recent user verification** — app/device recently unlocked with approved local authentication.
3. **Fresh biometric** — required immediately before a sensitive operation.
4. **Step-up/recovery** — second trusted device, recovery credential, or owner ceremony for root-level operations.

## Transaction authorization

For sensitive operations, sign the exact canonical action rather than a generic `approve=true` challenge.

Example:

```text
Approve:
  operation: deploy
  target: production-eu
  repository: OpenCoven/psyche
  commit: <sha>
  effect: modifies production
[...]

Issue #784, opened 2026-08-22
Plan: Identity-preserving mobile device pairing and trust architecture


BunsDev:
## Objective

Define the canonical OpenCoven trust model for connecting a phone to a local or remote Coven/Psyche node without collapsing owner identity, familiar identity, device identity, and session identity into one credential.

The target experience is: **introduce a device once, then reconnect seamlessly while retaining explicit, revocable authority boundaries.**

## Architectural principle

A phone is a **delegated device**, not the owner identity, Cave/Psyche identity, or familiar identity.

Trust chain:

`QR / trusted introduction → authenticated key exchange → device-bound credential → local user verification → scoped device grant → encrypted session`

### Credential separation

- **Owner/root identity** — issues and revokes delegations.
- **Familiar identity** — portable identity that survives runtimes and devices.
- **Device-bound key** — proves possession of one enrolled device.
- **Device grant** — expresses the authority delegated to that device.
- **Session key** — ephemeral encryption material for one connection.
- **Passkey/account credential** — optional recovery and remote-enrollment mechanism; not the familiar/root identity.

## Requirements

- [ ] Specify a versioned pairing/enrollment protocol and threat model.
- [ ] Define owner → installation/device → capability delegation semantics.
- [ ] Use pairwise/pseudonymous device identifiers rather than globally correlatable device IDs.
- [ ] Never transmit biometric templates, biometric hashes, hardware serial numbers, advertising IDs, or root private keys.
- [ ] Require proof-of-possession rather than reusable bearer credentials for enrolled devices.
- [ ] Define grant scopes, expiry, revocation, rotation, and recovery behavior.
- [ ] Bind security-sensitive approvals to the exact action/transcript being approved.
- [ ] Define behavior for compromised/lost devices and compromised rendezvous/relay infrastructure.
- [ ] Preserve self-hosted and open-source clients; platform attestation must not silently become the identity root.

## Proposed trust graph

```text
Owner Identity
    │
    ├── delegates → Coven/Psyche installation
    │
    └── delegates → Mobile Device
                       │
                       └── authorizes → sessions / actions / familiar access
```

## Security invariants

1. A relay can route traffic but cannot read or forge an authenticated OpenCoven session.
2. Stealing a stored grant without its private key is insufficient to authenticate.
3. A biometric is local user verification, never a network credential.
4. Enrollment is transcript-bound and resistant to substitution/relay attacks.
5. Revoking a device does not rotate or destroy a familiar's identity.
[...]

PR #791, opened 2026-08-23
feat: add versioned device grants to mobile authority

BunsDev:
## Summary

- adds an explicit Rust authority model for device scopes, assurance levels, transport restrictions, expiry, and revocation epochs
- defines canonical exact-action intents so sensitive approvals can bind operation, target, effect digest, nonce, and expiry
- atomically migrates the private device registry from v1 records to v2 records containing key-bound grants
- preserves the existing v1 pairing and request shapes while issuing deterministic grants behind the authority boundary
- re-checks grant identity, revocation epoch, active time window, and route scope before releasing a protected response
- adds migration, key-binding, scope, assurance, expiry, revocation-race, corruption, and privacy tests

## Compatibility

Existing paired devices migrate to a grant with exactly their existing `memory_read` authority. Migration does not widen permissions, rotate the device key, or change the mobile protocol payload. Revoked legacy devices migrate with an advanced revocation epoch.

## Security boundary

Biometrics remain a local platform gate on key use; this PR models the assurance requirement but does not accept an untrusted remote claim that a biometric occurred. Platform adapters and attestable step-up evidence remain follow-up work.

## Verification

The full workspace CI suite is required before merge. No dependency or lockfile changes are introduced.

Advances #786
Related architecture: #784

### DIFF
--- a/crates/coven-cli/src/mobile_memory/auth.rs
+++ b/crates/coven-cli/src/mobile_memory/auth.rs
@@ -9,7 +9,8 @@ use p256::ecdsa::{Signature, VerifyingKey};
 use sha2::{Digest, Sha256};
 use uuid::Uuid;
 
-use super::registry::{DeviceRecord, DeviceRegistry};
+use super::grant::{AssuranceLevel, DeviceScope};
+use super::registry::{DeviceAuthorizationRecord, DeviceRecord, DeviceRegistry};
 use super::MOBILE_REQUEST_WINDOW_SECONDS;
 
 const MAX_REPLAY_ENTRIES: usize = 10_000;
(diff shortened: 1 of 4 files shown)


BunsDev:
Follow-up hardening applied before merge:

- Enrollment capabilities now use the canonical v1 capability registry and reject unknown authority.
- The diagnostic `signature` field is no longer embedded in the EnrollmentRequest COSE payload; the COSE_Sign1 envelope is the authoritative signature, matching DeviceGrant, TransactionAuthorization, and RevocationRecord behavior.
- Deterministic vectors were regenerated and tests rerun after the change.

PR #839, opened 2026-08-28
docs(release): close the runbook postflight and tag-immutability gaps

BunsDev:
## Context

`coven-v041-runbook` requires the operator runbook to document: automated GitHub Release creation, bump-forward recovery from partial npm publication, tags never reused, and a postflight covering "packages, provenance, assets, checksums, installs, and the certification packet."

Audited `docs/reference/releasing.md` at `8ae39cd`. The first two were already documented thoroughly. Three requirements were not met:

1. **Installs were never verified.** Postflight checked `npm view` versions, provenance badges, and that the GitHub Release assets exist — but never installed the published packages. Nothing anywhere in the pipeline installs from the public registry: the `npm-onboarding` CI matrix only exercises locally packed tarballs. Dist-tag resolution and optional native-package selection — the two things most likely to break for a real user — were unverified at release time.
2. **The certification packet was absent from the runbook entirely**, despite being the artifact that ties a release to verified provider behaviour, and despite `coven-v041-cert` and `coven-v041-postflight` both depending on it existing.
3. **Tag immutability was stated only incidentally**, inside recovery procedures ("never moves or reuses the tag", "Never move or reuse a release tag"), never as a normative rule in the release-cutting flow where an operator would look first.

## Implementation

Docs-only, `docs/reference/releasing.md`, +96 lines.

- **Preflight -> Certification packet**: per-harness `coven setup <p> --verify-only --report-json`, with the `candidate_commit` equality check against the commit being tagged. Notes this is operator-only — verification needs a TTY and explicit network/cost consent, and a non-interactive shell reports `non_tty` and publishes nothing. Documents the exact redacted field set so operators know the packet is safe to retain.
- **Tag and push**: a normative immutability rule that explains *why* — the tag is the object npm provenance, the GitHub Release, and the certification packet all attest to, so moving it invalidates all three at once.
- **Postflight**: checksum verification of downloaded assets (`shasum -a 256 -c SHA256SUMS`), a new "Verify a fresh consumer install" section, and a "Confirm the certification packet" cross-check against `git rev-list -n 1 vX.Y.Z`.

The install section calls out the shadowed-global-install trap explicitly (`which -a coven` must return exactly one path), because a stale global install on `PATH` makes a broken release look fine.

## Verification

- `python3 scripts/check-api-contract-docs.py` -> rc 0
- `node --test scripts/onboarding-docs-test.mjs scripts/cli-docs-test.mjs` -> 23/23 pass
- `python3 scripts/check-secrets.py` -> rc 0
- `python3 scripts/check-coven-privacy.py --staged` -> rc 0

Every documented command and field was checked against the implementation rather than written from memory: the report schema against `SetupReport`/`publish_report` in `crates/coven-cli/src/setup/mod.rs`, the `--report-json` constraints against `docs/reference/cli-setup.md`, the `doctor` exit-1-with-guidance claim against `assertDoctorFailureGuidance`, and the install command against `README.md`.

## Risk

None to shipped code. The risk is documentation drift, which the existing docs tests cover.

## Agent Handoff

Unblocks `coven-v041-freeze` on the runbook axis. `freeze` still needs `coven-v041-cleanup`, and `coven-v041-npm` still needs authenticated Trusted Publishers evidence that cannot be produced from an agent session.

Note the dependency shape: this bead nominally depends on `coven-v041-npm`, but the npm preflight facts it would document are already recorded in that bead's notes. The only outstanding npm item is the Trusted Publishers settings evidence, which does not change any runbook text.

https://claude.ai/code/session_01J86YXtYgkNVZkDriuA3qJG

### DIFF
--- a/docs/reference/releasing.md
+++ b/docs/reference/releasing.md
@@ -82,7 +82,49 @@ You can leave the `npm-publish` environment itself in place or remove it — the
    ```
 4. Confirm the changelog and any README / brand updates have already landed on `main`.
 
+### Certification packet
+
+Produce the redacted certification report for every supported built-in harness
+against the exact commit you intend to tag. Each report is a one-provider
+artifact and the destination must not already exist:
+
+```sh
+coven setup codex   --verify-only --report-json cert/codex.json
+coven setup claude  --verify-only --report-json cert/claude.json
+coven setup copilot --verify-only --report-json cert/copilot.json
+```
+
+Verification requires an interactive terminal and explicit network/cost
+consent, so this cannot run in CI — it is an operator step on a machine with
+all three providers authenticated. A non-interactive shell reports `non_tty`
+and publishes nothing.
+
+Each report must show `"completed": true` and a `candidate_commit` equal to the
+commit being tagged:
+
+```sh
+jq -r '"\(.harness) \(.completed) \(.candidate_commit)"' cert/*.json
+```
+
+`candidate_commit` is baked into the binary at build time, so a report whose
+value does not match the release commit was produced by a different build and
+does not certify this release. Reports are success-only and published
+atomically; if verification fails, no file appears and the CLI exits nonzero.
+
+Keep the packet with the release record. It contains only `harness`,
+`cli_version`, `platform`, `candidate_commit`, `duration`, `exit_class`, and
+`completed` — no output, account data, tokens, or private paths — so it is safe
+to retain and share.
+
 ### Tag and push
+
+Release tags are **immutable**. A pushed release tag is never moved, deleted,
+recreated, or reused for different content, and a version that has been
+published to npm is never republished. Every recovery path in this document
+bumps forward to a new version instead. This is not a style preference: the
+tag is the object that the npm provenance attestation, the GitHub Release, and
+the certification packet all attest to, so moving it silently invalidates all
+three.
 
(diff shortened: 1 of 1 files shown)


BunsDev:
All four corrected in 68ccaff, and thanks — the second one was a genuine misreading of the code on my part, not just imprecise wording.

1. **`cert/` parent dir** — added `mkdir -p cert`. Confirmed `publish_report` uses `parent.join(...)` with `create_new(true)` and never creates directories.
2. **non-TTY behaviour** — you're right and my original text was wrong. On the `--report-json` path the `NonTty` result carries `version: None`, so `publish_report` hits `ok_or(SetupError::Rejected)?` and `run_setup` propagates the error *before* `render_human` runs. The CLI never prints `non_tty` there at all. Reworded to the observable behaviour: exits nonzero, writes no report.
3. **`which -a` portability** — split into labelled Unix and PowerShell (`Get-Command coven -All`) forms. Applied the same fix to the `SHA256SUMS` verification, which had the identical problem with `shasum -c`; the PowerShell loop is written against the generator's actual format (64 lowercase hex, two spaces, filename, verified in `package-github-release.mjs`).
4. **`npm ls -g`** — corrected. Now states the expected two entries (wrapper + the current platform's native package) and spells out which deviations are real release failures, plus that all five are never installed together on one machine.

Issue #816, opened 2026-08-24
Native familiar automations: replace harness-owned schedules with durable Coven routines


BunsDev:
## Program status — 2026-08-30

Parent roadmap: #854  
Tracker operationalization: #859

The native durable-routine foundation has **materially landed on `main`** through the recent automations implementation series. Current source and history show:

- Coven-owned versioned routine definitions;
- SQLite definition, occurrence, lease, and run storage;
- RRULE-backed daily/weekly and multiple-`BYHOUR` planning;
- unique occurrence fencing and bounded claim leases;
- expired-lease recovery, latest-only misfire behavior, and overlap refusal;
- daemon-side recurring tick and scheduled dispatch;
- a shared manual/scheduled runtime-launch path;
- familiar ID propagation into execution;
- bounded logs and atomic output delivery;
- health and run-history projections;
- source-preserving, paused legacy import;
- `coven.automations.*` control actions;
- Cave’s migration away from direct Codex ownership in OpenCoven/coven-cave#4990.

This issue therefore remains open for **foundation reconciliation and exact evidence**, not because the original architecture is still absent.

### Evidence required before closing #816

- [ ] Link the complete landed commit/PR series and identify the exact final foundation revision.
- [ ] Run and attach clean-clone verification for all automation tests rather than relying only on commit-message test counts.
- [ ] Verify schema migration from a pre-automations Coven database and prove rollback/recovery behavior.
- [ ] Prove daemon startup, immediate/recurring tick wiring, shutdown, and restart behavior on the supported local platforms.
- [ ] Prove one scheduled familiar-bound run and one manual run traverse the same claim, ledger, runtime, and delivery path.
- [ ] Prove a stale lease cannot block the next eligible occurrence indefinitely.
- [ ] Prove failed output delivery cannot be reported as success.
- [ ] Reconcile every original acceptance criterion below as implemented, intentionally deferred, or superseded.
- [ ] Link Cave’s completed ownership migration and document the remaining compatibility facade as downstream work.
- [ ] Add the foundation evidence packet to the Bead/GitHub mapping created under #859.

**Closing #816 will mean the original native local scheduler foundation is implemented and verified. It will not mean Coven Automations v1 is certified.**

### Remaining work has been split deliberately

The following outcomes are no longer implicit catch-all requirements for #816:

- #855 — stable `coven.automations.v1` schemas, state machines, command adoption/idempotency, typed failures, and changefeed.
[...]

Issue #857, opened 2026-08-30
P0: Bind automation runs to principal authority, familiar revisions, capabilities, approvals, and receipts


BunsDev:
Parent program: #854  
Protocol dependency: #855  
Foundation: #816

## Outcome

Ensure every Coven automation occurrence is dispatched only as an authenticated, explicitly authorized embodiment of the intended familiar, with fresh runtime capability checks, approval handling, replay resistance, and a verifiable receipt describing exactly what authority was exercised.

A routine carrying `familiarId = "charm"` is not sufficient proof that Charm ran, that the current principal authorized the run, or that a runtime had permission to perform its effects.

## Current gap

The landed foundation correctly propagates a familiar identifier into the shared runtime launch path and reuses Coven-owned execution state. However, the current automation definition and capability catalog do not yet make these guarantees normative:

- stable opaque principal identity and authenticated request authority;
- exact familiar root and identity revision binding;
- identity/declaration integrity digest;
- current authorization and revocation state at dispatch time;
- operation-specific capability grants and protected-surface decisions;
- approval requirements and approval consumption;
- runtime descriptor/capability compatibility;
- nonce, request-adoption, and replay protection;
- a tamper-evident run/authority receipt;
- privacy/retention handling for sensitive identity and authority evidence.

Until this issue is complete, recurring familiar work must be treated as a local execution convenience—not proof of authorized familiar continuity—and unattended external side effects must remain disabled or approval-gated.

## Canonical boundaries

This implementation must consume, not duplicate, canonical semantics:

- **Familiar Contract / continuity profile** defines familiar root, identity revision, same-familiar rules, and session embodiment binding.
- **Coven Threads / authority profile** defines protected actions, capability decisions, approval semantics, degrade-to-proposal behavior, and authority evidence.
- **Coven** resolves those inputs, authorizes dispatch, owns occurrence/run/attempt state, and commits the resulting receipt.
- **Psyche** may execute a multi-step authorized action but does not create identity or schedule authority.
- **Cave/SDK** propose, approve, inspect, and verify; they do not infer permission or author lifecycle truth.

## Required dispatch binding

Before runtime launch, atomically construct or reference an immutable `AutomationExecutionBinding` containing at least:

- automation ID and exact definition revision/digest;
- occurrence ID, canonical occurrence key, and current fence generation;
- run/attempt IDs and adopted request/idempotency key;
- authenticated principal ID;
- principal credential/key ID or authorization proof reference;
[...]

PR #872, opened 2026-08-30
feat(automations): truthful run settlement and output delivery (coven#816, part 9)

CompleteDotTech:
## Summary

Implements the next slice of Coven-native routine automations: **truthful run settlement, output delivery, and overlap enforcement** — the delivery/observability half of #816 (the "part 4" behaviors; parts 1, 5, 6, 7, 8 are already merged).

Before this change, a routine run was recorded `succeeded` with exit code `0` the moment `launch_session` returned — the ledger lied about runs that were still in flight or had failed. With this slice, Coven owns the whole lifecycle end to end:

- **Truthful settlement.** Dispatch leaves the run in flight: the occurrence moves to `running` under a bounded lease derived from the definition timeout, and the ledger row keeps `running` with the launched session id. A new reconciliation pass (`automations::delivery::settle_finished_runs`) settles both from the Coven session store — terminal status, exit code, and a bounded log of the normalized stream — and runs on every daemon tick and `coven.automations.tick`.
- **Atomic output delivery.** When a definition configures `outputTarget`, Coven (not the model) commits the final assistant payload with a tmp-file + rename atomic write. A failed output commit fails the run visibly; it is never reported as success. Nothing is delivered for failed runs.
- **Overlap enforcement.** `overlap: forbid` is now enforced at claim time: a routine with a live claimed/running occurrence rejects new claims (scheduled and manual alike). Run-now rejections fail visibly with an `overlap:` reason and record no ledger row, and never leave a stray planned occurrence behind.
- **Fail-closed wire defaults.** Definitions now default to `PAUSED`, runtime `coven-code`, misfire `latest`, overlap `forbid`, timezone `local` when fields are omitted on create.
- **Stale runs can never wedge a routine.** `running` rows carry bounded recoverable leases; lease recovery fails them with a recorded reason and the reconciliation pass settles their ledger rows.

`coven.scheduler` remains reserved for multi-host routing; recurring work stays advertised separately as `coven.automations`.

## Issue

Refs OpenCoven/coven#816

Refs OpenCoven/coven#816 (re-target upstream to close it)

## Test plan

Checked = ran locally on this host (no Rust toolchain available — cargo checks are deferred to CI):

- [x] `python scripts/check-secrets.py` — clean
- [x] `python3 scripts/check-coven-privacy.py --staged` — clean
- [x] Line-width audit (`<= 100 cols`) across all changed Rust files (rustfmt conformance)

Deferred to CI (no cargo/rustc on this host):

- [ ] `cargo fmt --check`
- [ ] `cargo clippy --workspace --all-targets -- -D warnings`
- [ ] `cargo test --workspace --locked`

New/updated focused tests: wire defaults; overlap-forbid claim rejection; bounded recoverable leases on `running` rows; dispatch leaves runs in flight (never instantly "successful"); overlap rejection records no second ledger row; bounded log keeps the tail within the 64 KiB budget; final-output selection; atomic output commit + temp-file cleanup; failed output commit fails the run visibly; failed session fails the run without delivery; recovered occurrence settles its ledger row; still-running rows untouched; daemon tick plans → claims → dispatches → settles; control-plane run-now/tick/runs end-to-end.

> **Vehicle note:** opened in the fork CompleteDotTech/coven as the CI vehicle — this token cannot write to OpenCoven/coven. Re-target upstream once write access is restored. Refs OpenCoven/coven#816.

---

> Recreated upstream from [https://github.com/CompleteDotTech/coven/pull/13], preserving source head `16d7b7f29df7125aa6434ddecab1313b66cf578a` and branch `agent/issue-816-native-familiar-automations-replace-harness-owned`.

### DIFF
--- a/crates/coven-cli/src/api.rs
+++ b/crates/coven-cli/src/api.rs
@@ -575,7 +575,8 @@ pub(crate) fn handle_request_with_runtime_and_authority(
                     );
                 }
             };
-            let (status, response) = control_plane::route_action(payload, &conn, runtime);
+            let (status, response) =
+                control_plane::route_action(payload, coven_home, &conn, runtime);
             json_response(status, &response)
         }
         ("POST", "/cast") => submit_cast(coven_home, body, runtime),
@@ -2106,7 +2107,6 @@ fn launch_session(
     // up, i.e. a child binding — an unbound or root-bound launch must not
     // create any store side effect ahead of the maintenance gate below, per
     // the existing maintenance-before-store-open precedence (§5.1).
-    let mut opened_conn: Option<rusqlite::Connection> = None;
     if let Some(binding) = execution_binding.as_ref() {
         let Some(familiar) = familiar_ctx.as_ref() else {
             return api_error(
@@ -2139,7 +2139,6 @@ fn launch_session(
             {
                 return Ok(response);
             }
-            opened_conn = Some(conn);
             // Test-only failpoint (issue #728 BLOCKER 1, final review): a
             // no-op in every non-test build. Deterministic concurrency tests
             // hook this to run a real concurrent write — e.g. sacrificing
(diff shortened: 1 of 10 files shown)


BunsDev (changes requested):
The direction is right, but settlement is not yet truthful or durable. Blocking findings:

1. Automation launches call the runtime directly and bypass durable session-row/event creation plus the normal maintenance/familiar checks. Extract one durable launch primitive shared with `/sessions`; persist occurrence/run/session/snapshot/lease before spawn and terminally settle every launch failure.
2. `coven.automations.tick` claims work but never dispatches it. Later daemon ticks only dispatch newly claimed occurrences, so control-plane claims and crash-interrupted claims can remain stuck forever. Share a full tick implementation or make the action planning-only, and always scan valid existing claims for dispatch.
3. Timeout recovery can leave occurrence=`failed` while run=`succeeded`, and timeouts above 1440 minutes are silently truncated despite validation allowing more. Reconcile one authoritative terminal decision transactionally using actual completion/deadline, surface CAS failures, and align supported timeout bounds.
4. Settlement reloads the mutable current definition, so editing/deleting `outputTarget` in flight changes delivery semantics. Pin immutable definition revision/digest and delivery target on claim/run.
5. Delivery selects only the last raw output chunk. Persist a canonical final-assistant event or reconstruct all ordered chunks without the log-tail limit.
6. Latest-only misfire scanning stops after 96 slots and can schedule months-old work after long downtime. Compute the true latest slot or return an explicit degraded error.
7. Until #857 lands, restrict automation mutations/execution to owner-local IPC. TCP is not owner-authenticated, `/actions` currently ignores authority, and the docs overstate fresh authority/approval binding.

Also address crash durability/idempotency around file delivery, collision-prone millisecond IDs, and the deferred manual retry/attempt policy before calling this end-to-end settlement.

Issue #971, opened 2026-09-09
ci: isolate native dependency installation from unrelated runner repositories


BunsDev:
Linux Rust lint/tests and onboarding are blocked before compilation by a repeated Google Chrome apt index SHA-256 mismatch on the Ubuntu runner. Reproduced on #970 (853e778, run 34384340759 attempts 1 and 2) and #933 (b7d9bc11, run 34384351627 attempts 1 and 2). Windows jobs pass. The requested package is libopenblas-dev from Ubuntu, not Chrome. Scope: narrow native dependency installation to authenticated distribution sources, reuse one coherent helper where appropriate, preserve signature/hash verification and fail-closed errors, and prove actual Linux CI green. Do not ignore apt failures, weaken integrity, remove required gates, or alter global runner repositories. Related Threads execution epic: threads-8pz.

PR #973, opened 2026-09-09
ci: isolate native dependency installation to authenticated Ubuntu sources

BunsDev:
## Objective

Fix the native-link dependency CI blocker tracked in #971. Two attempts each on
#970 and #933 failed before Rust compilation because the runner's unrelated
Google Chrome package index did not match its signed release hash.

## Change and acceptance

Use one helper for the existing eleven native dependency installs across CI,
npm release, and release-stress workflows. Both update and installation use
only the expected Ubuntu distribution source and isolated package lists.
Deb822 sources, hosted-runner mirror files, and legacy distribution lists are
supported. Existing signature/hash checks remain enabled; missing expected
source/signing information and apt failures remain hard errors. The helper
does not remove or modify the runner's global repositories. Its public metadata
directory remains traversable by apt's unprivileged downloader, and package
arguments cannot become apt flags.

## Evidence and impact

The no-sudo Python regression suite covers source scoping, hosted mirror
configuration, legacy filtering, missing signing metadata, and propagation of
update/install failures. Existing workflow, classifier, secret, and staged
privacy guards passed. Hosted Linux installation/build results remain pending
on this exact checkpoint; local macOS stubs are not represented as that proof.

No runtime authority, credential, database schema, or migration semantics
change. Release jobs keep their required gates. Reverting this helper/call-site
change restores the previous installer without changing runtime artifacts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,6 +111,7 @@ jobs:
       - run: python3 scripts/check-reliability-scorecard.py
       - run: python3 scripts/check-workflows-test.py
       - run: python3 scripts/check-ci-workflow-test.py
+      - run: python3 scripts/install-native-link-dependencies-test.py
       - run: node --test scripts/package-github-release-test.mjs
       - run: node --test scripts/package-automations-protocol.test.mjs
       - run: node --test scripts/package-automations-authority-profile.test.mjs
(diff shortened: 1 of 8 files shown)


BunsDev:
The exact eight-file #971 fix from this PR is already included in #970 (verified byte-for-byte). I am finishing #970 through protected merge, including that prerequisite, and will close this duplicate only after confirming all eight files landed unchanged. No work from this PR is being discarded.

BunsDev:
Superseded by merged #970 (4a04dc927f5ab1ba3d6a725373cf0e477eefd386), which closes #971. All eight files from this PR were included byte-for-byte before the final review hardening. The only subsequent prerequisite changes add complete Ubuntu URL matching and per-stanza/per-URI validation with regression coverage. Verified the final #970 head and squash-merge trees are identical. This PR is closed as superseded, not represented as separately merged.

Issue #977, opened 2026-09-09
Bind approved proposal commit to the same revalidated authority snapshot


BunsDev:
## Completion: landed and checked off

The snapshot-binding correction from #979 is preserved on main through #1022
and the later integrated checkpoints. The pre-merge-only status below is
historical.

- [x] Reproduce the original defect deterministically at the committed pre-fix regression source.
- [x] Bind applying intent to the actual validated baseline snapshot and candidate identity commitment, rather than independently rereading a new snapshot during intent construction.
- [x] Recheck live Ward configuration, the same resolved baseline keys, and authoritative identity evidence around conditional writes.
- [x] Refuse changed/unavailable authority while preserving concurrent bytes and distinguishing proven rollback from ambiguous application.
- [x] Preserve typed terminal semantics and separately report cleanup failures after proven rollback.
- [x] Exercise the feature-gated real-daemon final-commit drift journey and negative recovery cases.
- [x] Integrate on main with retained local and native Windows acceptance.

### Original red and fixed green

The original test-first commit
`596cb7a7347653c35a376b50797b0093995d5e7a` (parent `299afacf`) was replayed
without source edits in an isolated detached worktree:

```sh
cargo test --locked -p coven-cli --bin coven --   api::tests::identity_predicate_approve_refuses_identity_drift_at_final_commit   --exact --nocapture
```

Exactly one test ran and failed behaviorally: the unchanged pre-fix code returned
HTTP **200** with `decision=approved` after the injected identity drift, rather
than the asserted HTTP **409**. This was not a compilation error or zero-test
result. The raw receipt is retained with SHA256
`c12e8d8490044d06f7c2381176a09210e5e84c79f1d957d81c1e3286ee8e1756`.

The same named regression passes in the retained `bdd9ea20` workspace and native
Windows receipts. The real-daemon
`final_commit_identity_drift_closes_opened_window_without_applying` case passes
on local macOS and in [Windows CI 34733723336](https://github.com/OpenCoven/coven/actions/runs/34733723336).
Unavailable/diverged recovery, proven-rollback restoration, cleanup-failure
typing, and concurrent-byte preservation regressions also pass.

That Windows run executed synthetic `ee1d0a7e` with the exact `c77eba4f` tree
that landed as main `c017f60d`. Artifact `10309938925` is retained with SHA256
`97ab1ebc5cdba0adad8795237fa6de99fcf2e3c66d76fe76f8cf7c63c1c7a987`.
[...]

PR #931, opened 2026-09-06
test(threads): integrate real-daemon authority and replay journeys

BunsDev:
## Current integration checkpoint: 8576f41

Artifact names now include the commit, workflow run, and attempt, matching the
already-isolated storage roots. Two diagnostic attempts at the same commit
previously produced identical names; both were preserved by exact artifact ID.
The naming change prevents ambiguous name-based selection. Thirteen workflow
guards pass; no daemon or harness behavior changed. Native CI `34489425644`
passed, with 20/20 Windows target cases in 32.20 seconds and 152 selected
feature-unit cases. Artifact `10157780426` has a commit/run/attempt-qualified
name and 28 passed manifests, all clean merge
`0d8ccfca1725475b5e45f2ec47d8c4c2c45bc89d`. #1000 is implemented and verified
in this draft, but remains open for reviewed main integration.

## Previous measurement checkpoint: 92b701a

#1001 adds bounded advisory remaining-budget samples around start/restart launch
and fixed store checkpoints for connection configuration, Ward, runtime schema,
main lock, main schema, and commit. Ordinary store initialization uses a no-op
observer. SQL, transaction ownership, durability, deadline accounting, retries,
cleanup, and authority semantics are unchanged.

Parent review and 37 selected lifecycle/store/logging regressions passed.
The complete 20-case target passed at the committed head in 18.16 seconds with
isolated artifacts and the committed Threads dependency. Native CI
`34485088291` attempt 1 passed, including 20/20 Windows target cases in 40.18
seconds. Preserved artifact `10156032323` has 28 passed manifests, all merge
`bec6c2bdcc951c1ba8e19d9dfb27feb1f0c48bc6`, with 63 complete phase/budget
sequences. Before-launch remaining budget ranged from 1935 to 1995 ms;
after-launch ranged from 1928 to 1992 ms. The maximum cumulative
daemon-store observation was 1327 ms. These successful samples do not diagnose
the earlier failures. The one additional diagnostic Windows attempt also
passed. Its separate artifact `10156935220` has 28 passed current-merge
manifests and 63 complete sequences: remaining budget was 1946-1994 ms before
launch and 1940-1990 ms afterward, with a maximum cumulative daemon-store
observation of 653 ms. Neither attempt reproduced the failure. Deliberate
diagnostic repetition stops here; retained instrumentation can capture a
future occurrence. Neither green attempt resolves startup causation or proves
the first-attempt stability requirement.

The independently proven macOS socket-leaf repair landed on main through
[...]

PR #979, opened 2026-09-10
fix: bind approved proposal commits to validated authority

BunsDev:
## Context

Bind the final approved Threads commit to the exact authority evidence that was revalidated. Addresses #977 and stacks on #931; it does not introduce a new identity model, audit store, or cross-file transaction guarantee.

## Implementation

- Persist the exact Gate-2-resolved `ward_manifest` baseline snapshot used to build the validated weave, rather than rereading requested target keys while creating applying intent.
- Bind that snapshot, Ward config, proposal authority bytes, and validated candidate identity commitment into recovery commitment v3.
- Re-read the same resolved baseline keys immediately around conditional Ward writes; authority drift fails closed and rolls back owned target changes.
- Distinguish unknown recovery (`NoWrite`/`Ambiguous`) from proven rollback. Proven rollback restores ordinary proposals for fresh validation or terminally closes an already-opened window.
- Preserve typed authority-drift causes through rollback cleanup failures and report leftover temporary-artifact targets separately.
- Add deterministic unit coverage plus a feature-gated real-daemon pause journey.

## Acceptance and verification

- [x] 14 identity-predicate/commit-binding unit regressions pass.
- [x] All 16 `threads_e2e` journeys pass with `threads-test-clock`, including final-commit identity drift.
- [x] `cargo fmt --check`.
- [x] `cargo clippy --workspace --all-targets -- -D warnings`.
- [x] Secret and staged privacy guards.
- [x] Independent review findings repaired and re-review clean.
- [x] Full workspace reached 2,985 passing tests; one unrelated daemon lifecycle timeout failed, then passed three isolated retries. Two earlier full runs likewise reached unrelated smoke-daemon races that passed three isolated retries each.

## Risk and rollback

This is an authority-boundary and recovery-semantics change. Older applying records without the persisted baseline snapshot fail closed as unverifiable rather than reconstructing unvalidated evidence. Ambiguous writes remain durable; only proven rollback may restore or terminalize the proposal. Roll back the six commits in this stacked PR together if the dependency checkpoint changes.

### DIFF
--- a/crates/coven-cli/src/api.rs
+++ b/crates/coven-cli/src/api.rs
@@ -10966,6 +10966,11 @@ fn decide_threads_proposal_inner(
                 }),
             );
         }
+        if let Err(error) = pause_threads_final_commit_fixture(coven_home) {
+            claim.preserve();
+            audit_reservation.preserve()?;
+            return Err(error);
+        }
         let (report, apply_cleanup_error) = match apply_after_review_approval(
             &ward,
             review_kind,
@@ -11282,6 +11287,11 @@ fn decide_threads_proposal_inner(
             audit_reservation.finish()?;
             return Err(error);
         }
+    }
+    if let Err(error) = pause_threads_final_commit_fixture(coven_home) {
+        claim.restore_pending(&document)?;
+        audit_reservation.finish()?;
+        return Err(error);
     }
     let expected_before = proposal_expected_before(&applying)?;
     let (report, apply_cleanup_error) = match apply_after_review_approval(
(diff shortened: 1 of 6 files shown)


BunsDev:
Final checkpoint is green at `d5a0a5a`. All applicable GitHub checks passed, including Linux and Windows Rust tests. The independent re-review confirmed all three findings are resolved: exact validated baseline binding, proven-rollback recovery handling, and typed cleanup-failure terminal semantics.

PR #1089, opened 2026-09-15
Restrict external harness manifests to approved adapter contracts

BunsDev:
### Motivation

An external adapter manifest declares `executable` plus argv prefixes, and Coven appends the user's prompt as the trailing argument. `ExternalHarnessAdapterSpec::into_spec` validated the executable's *shape* — no slashes, no whitespace — but never its **identity**. A manifest declaring:

```json
{ "id": "pwn", "executable": "sh", "non_interactive_prompt_prefix_args": ["-c"] }
```

passed every check and turned each prompt into `sh -c <prompt>` — arbitrary command execution from prompt text.

**Scope of the hole.** The trusted `COVEN_HOME/adapters` path was already safe: it is pinned by byte-equality against a shipped recipe and executes the in-memory recipe, which `configured_harness_specs_ignore_coven_home_adapter_manifest_that_differs_from_recipe` already proves. The live hole was the *arbitrary* path — manifests supplied via `COVEN_HARNESS_ADAPTER_MANIFEST` / `COVEN_HARNESS_ADAPTER_DIRS`.

### Description

Two checks in `into_spec`, both scoped to that concern:

- **`executable` may not name a shell or language interpreter** — `sh`, `bash`, `zsh`, `python`, `node`, `perl`, `env`, `powershell`, … — matched case-insensitively and ignoring a `.exe` suffix, so `SH`, `Python3`, `node.exe` and `PowerShell.EXE` are all rejected.
- **Neither prompt-prefix vector may contain a prompt-evaluating flag** (`-c`, `-e`, `--eval`, `--command`, `/c`). Those are precisely the argv the prompt is appended to.

**Why this keys on the executable rather than banning argv tokens.** `-c` is legitimate argv for non-interpreter runtimes: Codex's config-override form is `model_arg_template: "-c model={model}"`, and the `templated` fixture on `main` depends on it. A blanket token ban would have broken working adapters. A non-interpreter executable does not evaluate its arguments, so banning the interpreter is both sufficient and non-breaking.

Docs updated in `docs/HARNESS-ADAPTERS.md` and `docs/harnesses/hermes.md` to state both rules and the `-c` carve-out.

### Note on this branch's history

This PR was rebased from a base **1040 commits stale** (2026-06-06). In the interim `harness.rs` grew from 960 to 6,264 lines and the adapter architecture gained a recipe system (`grok`, `hermes`, `opencode`) plus the `TrustedRecipe` byte-equality trust model. The original commit's approach — a hardcoded allowlist containing exactly one entry (`hermes`), enforced for every manifest source — could not be applied to current `main`: it would have rejected `grok` and `opencode`, deleted the documented external-manifest feature, and broken 16 non-recipe adapter fixtures in `main`'s own tests.

The fix was therefore reimplemented against the current architecture. The vulnerability it addresses is unchanged and was re-verified as live on `main`. #1114 (a duplicate pointing at the identical commit) is closed; the original commit `f23e0820` remains at `refs/pull/1114/head`.

### Testing

All `AGENTS.md` CI gates run locally, green:

| gate | result |
|---|---|
| `cargo fmt --check` | pass |
| `cargo clippy --workspace --all-targets -- -D warnings` | exit 0, 0 warnings |
| `cargo test --workspace --locked` | exit 0, **0 failures** (3,489-test `coven-cli` suite included) |
| `python3 scripts/check-secrets.py` | pass |
| `python3 scripts/check-coven-privacy.py --staged` | pass |

No npm/TypeScript packages touched.

Four new regression tests, plus confirmation that the 122-test `harness::` module still passes in full — no existing fixture regressed:

- `external_manifest_rejects_interpreter_executable` — `sh` is refused
- `external_manifest_rejects_prompt_evaluating_prefix_args` — `-c` in a prompt-prefix vector is refused
- `external_manifest_rejects_interpreter_executable_case_and_exe_suffix` — `SH`, `Python3`, `node.exe`, `PowerShell.EXE`
- `external_manifest_still_allows_model_arg_template_config_flag` — guards the `-c model={model}` carve-out against future over-tightening


BunsDev:
Rebased onto current `main` (`2b72ed90`). Force-pushed `f23e0820` → `d9b05063`; the original commit is preserved at `refs/pull/1114/head`.

This was not a mechanical rebase, and the PR body has been rewritten to match what actually landed. Flagging the substantive change for reviewers:

The original commit's approach was a hardcoded allowlist with **one entry** (`hermes`), enforced in `into_spec` for every manifest source. Its base predates 1040 commits of drift — `harness.rs` went from 960 to 6,264 lines, and the adapter layer gained a recipe system (`grok`, `hermes`, `opencode`) and the `TrustedRecipe` byte-equality trust model. Applied verbatim it would have rejected two adapters `main` ships recipes for, removed the documented `COVEN_HARNESS_ADAPTER_MANIFEST` extensibility, and broken 16 non-recipe fixtures in `main`'s own tests.

So the guard now keys on **executable identity** (no shells or interpreters) plus prompt-evaluating flags in the two prompt-prefix vectors, rather than an identity allowlist. That closes the same `sh -c <prompt>` hole while leaving the extensibility path intact.

One deliberate carve-out worth reviewing: `-c` is **not** banned globally, because Codex's config-override form `model_arg_template: "-c model={model}"` is legitimate and `main`'s `templated` fixture depends on it. A non-interpreter executable does not evaluate its arguments, so the interpreter ban is what carries the security weight. `external_manifest_still_allows_model_arg_template_config_flag` pins that carve-out so it is not tightened away by accident.

Also confirmed while rebasing: the `COVEN_HOME/adapters` trusted path was **already** protected by byte-equality — the residual hole was only the env-var manifest path, and that is what this now closes.

BunsDev:
CI note: the `Rust tests (Windows)` failure on `ea1747dd` is a pre-existing flake, not this change. Filed as #1116.

`failure_artifacts_retain_failed_restart_cli_output` timed out on daemon startup health (`child=still-running ... last_readiness=connect-timeout`). In the **same run** (35156192481), the dedicated `Threads daemon tests (Windows)` job ran the same test from the same bytes and passed — so it is runner load, not logic. `Rust tests (Linux)` also passed.

This PR's diff touches `crates/coven-cli/src/harness.rs` and two docs; `threads_e2e.rs` has no reference to adapter manifests, `into_spec`, or the new guard.

`Policy guard` failed on the first attempt for a real reason and is now green: `docs/harnesses/hermes.md` needed a `source_adjacent_reason` frontmatter field (it was already non-compliant on `main` — the guard only inspects files a PR touches, so editing it surfaced a latent gap).

PR #1090, opened 2026-09-15
fix(cli): harden coven-code delegation lookup

BunsDev:
### Motivation
- The interactive delegation path previously scanned every `PATH` entry and `exec`ed the first `coven-code` executable it found, which allowed attacker-controlled repo-local binaries (empty/relative `PATH` components or `.`) to be selected and run with the user's privileges.
- The change hardens helper discovery so the CLI will not automatically run untrusted, workspace-local, or relative helper binaries while preserving the intended delegation behavior for legitimate installs.

### Description
- Skip empty and non-absolute `PATH` entries when searching for `coven-code` and prefer only absolute entries to avoid resolving into the current working directory.
- Canonicalize candidates and validate they are executable, that they are not inside the current workspace, and (on Unix) that every ancestor directory is owned by the effective user or root and is not group/world-writable before accepting a candidate.
- Added helper functions `coven_code_binary_name`, `is_trusted_coven_code_binary`, `has_trusted_ownership_and_permissions`, `is_in_current_workspace`, and `nearest_workspace_root` and integrated them into the lookup flow.
- Added regression tests covering empty `PATH`, relative project-local `PATH` entries (e.g. `node_modules/.bin`), and absolute helper paths inside the current workspace.

### Testing
- Ran `cargo fmt --check` (succeeded after formatting changes were applied).
- Ran the focused unit tests `cargo test -p coven-cli coven_code_binary_lookup -- --test-threads=1`, which passed (all added lookup-related tests succeeded).
- Ran `cargo test -p coven-cli`, which executed the crate test suite and showed the new unit tests passing but reproduced an unrelated existing smoke test failure (`smoke_daemon_session_replay_and_safe_session_rituals`), where the test environment reported a daemon PID did not exit after `SIGTERM`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a263f54d41483319d6cd3680a442089)


BunsDev:
Closing without merge; the fix is re-filed against the code that actually resolves the engine now.

The function this patches, `coven_code_binary()`, is a one-line shim over `engine::resolve()` on `main` (`main.rs:2286`); the PATH scan lives in `engine::resolve_from` step 3 (`engine.rs:120-133`), so this branch conflicts and does not touch the real lookup. The underlying point stands — that scan still honours empty and relative `PATH` entries — and is fixed by skipping non-absolute entries there, with a test in `engine.rs`.

The ownership walk is not ported on purpose: `mode & 0o022` rejects group-writable directories, which is the default on user-private-group systems (Debian/Ubuntu `umask 002` creates `~/.local/bin` and `~/.coven-code/bin` as 775), so it would report the engine as missing on stock installs. The tests also mutate process-global cwd and env with no lock.

PR #1096, opened 2026-09-15
fix: verify npm release tag before repository-controlled release code

BunsDev:
### Motivation

- Prevent a release-gate bypass where repository-controlled code (`scripts/release-npm-context.mjs`) could run from an attacker-pushed tag before the signed/annotated tag and `origin/main` ancestry checks completed.

### Description

- Move the `Derive stable or recovery release context` step (`node scripts/release-npm-context.mjs`) so it runs after local tag annotation, GitHub signature verification, allowed-signer `git verify-tag`, and `origin/main` ancestry checks in the `verify-tag` job in `.github/workflows/release-npm.yml`.
- Preserve recovery-mode validation by performing the recovery-specific base-tag verification and release-only file-diff checks in a dedicated post-verification step that runs only when `release_mode == 'recovery'`.
- Add a regression assertion to `scripts/publish-npm-test.mjs` that ensures `git verify-tag` appears before any invocation of `node scripts/release-npm-context.mjs` in the release workflow text.
- Keep the rest of the release gate logic intact so the job still enforces allowed signers, GitHub verification, and release-only diffs for recoveries.

### Testing

- Ran `node --test scripts/publish-npm-test.mjs` and all tests passed (`57` tests passed).
- Verified workflow YAML parses with `ruby -e "require 'yaml'; YAML.parse_file('.github/workflows/release-npm.yml')"` and it is valid.
- Ran secret-scanner suites with `python3 scripts/check-secrets-test.py` and `python scripts/check-secrets.py`, both passed (secret guard tests passed and no findings).
- Ran `python3 scripts/check-coven-privacy.py --staged` on the changes and it passed for staged files.
- Ran `cargo fmt --check` and it succeeded; a full `cargo clippy && cargo test` attempt was started but the Rust build was blocked by an external dependency download failure (`ort-sys` prebuilt binary download returned HTTP 403), preventing completion of compilation-based checks in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a6d9eeded608331a5c33e8ba632ce17)


BunsDev:
Closing without merge. Two reasons:

1. The threat model does not hold for `on: push: tags`. The workflow YAML itself is taken from the tagged commit, so whoever can push a `v*` tag already controls every step, and the parallel `verify` job (no `needs`) runs `cargo test` on that commit before any gate. Moving the 30-line `release-npm-context.mjs` parser after `git verify-tag` changes nothing about what tag-push access can do. The real controls — tag-push permissions, `NPM_RELEASE_ALLOWED_SIGNERS`, and OIDC on `npm-publish` behind `needs: verify-tag` — are untouched.
2. It re-introduces the recovery-release path that #734 (`35ed0abf`) removed as unsafe; `scripts/publish-npm-test.mjs` on `main` asserts `release_mode` and the recovery plumbing do not reappear, so this branch fails the suite it edits. It also conflicts in `release-npm.yml` (722 commits behind).

If the cosmetic reorder is still wanted it is a three-line move of the `release-context` step on top of current `main`, not this branch.

Issue #1084, opened 2026-09-15
Terminal UI: wire the theme layer into unstyled surfaces, add depth, and amend the Cast contract for a launcher masthead


BunsDev:
## Context

A review of the terminal UI surfaces found that `crates/coven-cli/src/theme.rs` is a well-built brand layer — brand tokens mirrored from `brand/ui/color-tokens.css` with a drift test, truecolor/256/NoColor detection, and a real nearest-256 downgrade — that is **not reaching most of the surfaces users actually look at**.

Measured on `main` (375c8032), with `CLICOLOR_FORCE=1 COLORTERM=truecolor`:

```
coven doctor            escapes=0
coven status            escapes=0
coven sessions --plain  escapes=0
```

`theme::Status::{Working, Warning, Error, Idle}` and `theme::Bg` are all `#[allow(dead_code)]`.

## Scope

Six items, tracked here, landing as separate scoped PRs.

### Contract-legal (no amendment needed)

`docs/design/cast-tui-contract.md` §1 scopes itself to the launcher and Cast cards. The chat TUI is explicitly "out of scope — separate visual track"; the session browser is "later phase"; doctor/status output is "CLI prose, not surfaces". These five items sit in that unclaimed space.

1. **Wire the theme into the CLI prose surfaces** — `doctor`, `status`, `sessions --plain`. Use the existing `Status` semantic. Also fixes defects found in the same output: `coven-code` (10 chars) overflows the 8-wide HARNESS column and shears every column after it; a title containing a newline splits a row across two lines; titles hard-cut with no `…`.
2. **Code blocks get a surface** — `tui/chat/render.rs:305-322` renders fenced code with a `  │ ` gutter and foreground tints only, so it bleeds into surrounding prose. Add a `SURFACE_2` fill to wrap width plus a mono uppercase language chip. Same area: `render.rs:306` **truncates** code lines instead of wrapping, silently losing code off the right edge.
3. **Streaming gutter** — spend the tick budget already paid during `is_responding` (`tui/chat/app.rs:329`, 120ms) on a ramped progress gutter with harness, token count, elapsed. Requires an `Rgb` lerp in `theme.rs`; none exists. Must not animate when idle — `tick_timeout()` returning `None` when idle is deliberate (commit 9f9a013d, "perf: block idle chat event loop").
[...]

PR #1102, opened 2026-09-15
fix(status): report the real build version instead of the 0.0.0 placeholder

BunsDev:
## The bug

```
$ coven --version
coven v0.4.2 (engine coven-code 0.7.0, pinned 0.7.0)

$ coven status
  version    0.0.0          <-- wrong
```

## Root cause

**`crates/coven-cli/Cargo.toml` pins `version = "0.0.0"` permanently** — Coven's real
version comes from the release tag, not the manifest. `build.rs` already resolves it
(`$COVEN_BUILD_VERSION` → `git describe --tags` → `$GITHUB_REF_NAME`) and stamps it into
`COVEN_VERSION_DESC`, which is what `coven --version` prints.

But the health payload's `covenVersion` read the manifest instead:

```rust
// crates/coven-cli/src/api_health.rs
pub const COVEN_VERSION: &str = env!("CARGO_PKG_VERSION");   // always "0.0.0"
```

So **every Coven build ever made has reported `covenVersion: "0.0.0"`.** The CLI was
printing exactly what it was told.

### It was never a stale daemon

Two independent confirmations:

1. `coven status` renders `/api/v1/health` through the **in-process** API router
   (`observe.rs` → `api::handle_request`), not over the socket. The `0.0.0` and the
   `v0.4.2` in the report came out of **the same binary** disagreeing with itself.
2. On the wire too — a current daemon (advertising `afsCommitDryRun`,
   `executionBindingContracts`, `requestAdoptionContracts`) answers
   `curl --unix-socket ~/.coven/coven.sock .../api/v1/health` with `"covenVersion":"0.0.0"`.

## The fix

**Root cause** — `api_health::coven_version()` now derives from `COVEN_VERSION_DESC`, the
same build stamp `--version` uses, so the two can no longer disagree. Normalization:

- the tag's `v` prefix is dropped, so the field stays a version string;
- `git describe` distance/commit/`-dirty` suffixes are **kept** — they *are* the build identity;
- when nothing resolved a version, it reports `unknown` rather than republishing the
  placeholder as though it were a real release.

**Defensive accompaniment** (clearly separate) — `observe.rs::status_version` maps a
`0.0.0`, blank, non-string, or absent `covenVersion` to `unknown`. This matters because a
daemon built *before* this change still answers health with the placeholder; without it,
nothing improves until the daemon is rebuilt.

## Verified end to end

```
$ ./target/debug/coven --version
coven v0.4.4-6-g3eb633b2-dirty (engine coven-code 0.7.0, pinned 0.7.0)

$ ./target/debug/coven status
  version    0.4.4-6-g3eb633b2-dirty

$ ./target/debug/coven status --json | grep covenVersion
    "covenVersion": "0.4.4-6-g3eb633b2-dirty",
```

## JSON contract: no shape change

`coven status --json` and `/api/v1/health` keep every key and type. Only the **value** of
`covenVersion` stops being a placeholder. That field is already specified as opaque daemon
build identity (`docs/OPERATIONAL-MODEL.md`: *"retain `covenVersion` in health as daemon
build identity distinct from the named contract version"*), and nothing parses it as
semver — `coven-client` requires only that it be non-empty, and the openclaw bridge passes
it through untouched. Compatibility negotiation remains on `apiVersion` + `capabilities`.

Docs were carrying `"covenVersion": "0.0.0"` in their health examples, teaching the bug as
normal; those now show a real version, plus one sentence in `docs/API-CONTRACT.md` saying
what the field actually is.

## Tests

- `api_health::build_identity_normalizes_every_descriptor_shape` — every descriptor shape:
  `v0.4.2`, `0.4.2-14-g3eb633b2-dirty`, a bare `--always` sha, the no-git fallback, blank.
- `api_health::health_version_is_never_the_manifest_placeholder` — pins the root cause:
  the health body must not report `0.0.0` or `CARGO_PKG_VERSION`.
- `observe::render_status_reports_placeholder_version_as_unknown` — the rendered line must
  not read `0.0.0` for a placeholder, blank, null, or absent field.
- `observe::render_status_passes_through_real_versions` — real identities render verbatim.

## Known same-root-cause spot left alone

`executor_node.rs::build_probe` sets the executor probe's `covenVersion` from
`env!("CARGO_PKG_VERSION")` — the identical defect on a different route. **I could not
include the one-line fix:** staging that file trips the `check-coven-privacy.py`
pre-commit hook on two *pre-existing, untouched* lines (`/home/coven/.ssh/id_ed25519`, an
SSH argv test fixture at lines 927/946, present on `origin/main`). The hook scans whole
staged files, so `executor_node.rs` is currently uncommittable without first clearing that
false positive. Flagging rather than bypassing the hook — worth a follow-up.

## Gates

| Gate | Result |
| --- | --- |
| `cargo fmt --check` | pass |
| `cargo clippy --workspace --all-targets -- -D warnings` | pass |
| `cargo test --workspace --locked` | 3440 passed; only the 3 known environmental `memory_import` `SUN_LEN` failures (long worktree path) |
| `python3 scripts/check-secrets.py` | pass |
| `check-coven-privacy.py` (`--staged` and `--range`) | pass |
| `check-docs-ownership.py --range origin/main...HEAD` | pass |
| `python3 scripts/check-api-contract-docs.py` | pass |

Rebased on current `origin/main` (3eb633b2). #1099 (`observe.rs` color) has **not** landed;
this diff touches only the version line and adds one helper above `render_status`, so the
plain/styled split should merge cleanly either way.

Refs #1084


BunsDev:
Added the deferred half in `bbcccc72..HEAD`: the executor probe now reports the real build identity too.

`build_probe` set `covenVersion` from `env!("CARGO_PKG_VERSION")` — the same pinned `0.0.0` — so every executor probe told the hub `0.0.0` regardless of which build answered. It now calls `api_health::coven_version()`, so one build reports one identity on both `/api/v1/health` and the probe.

**On the blocker that deferred it:** `executor_node.rs` carried `/home/coven/.ssh/id_ed25519` in an SSH argv fixture, which matches the privacy guard's home-path pattern. The guard scans whole staged files, so those two pre-existing lines made the file uncommittable for anyone.

I did not weaken the pattern — it is doing its job. The fixture is about argv layout, not about that particular path, so it now uses `/etc/coven/executor_id_ed25519`: still an absolute identity-file path exercising `-i <path>`, but not shaped like a real user's home directory. The file is now editable again.

New test `probe_reports_the_real_build_identity_not_the_manifest_placeholder` pins both properties: the probe never reports `0.0.0`, and it agrees with `/api/v1/health`.

Rebased onto current `main` (branch protection requires up-to-date), commits still signature-verified.

BunsDev:
Rebased onto `main` after #1099 landed; the two touched the same lines in `render_status` and it did not merge cleanly, contrary to my earlier note on this PR.

**Conflict:** #1099 replaced the `daemon`/`version` rows with palette-aware output (`field(p, …)`, `p.status(…)`), while this PR replaced the inline `covenVersion` lookup with `status_version(health)`.

**Resolved by taking both** — #1099's styled rows, sourcing the value from this PR's helper:

```rust
out.push_str(&format!(
    "{}    {}\n",
    field(p, "version"),
    status_version(health)
));
```

So a placeholder version now renders as `unknown` *and* picks up the muted field label, rather than one change silently reverting the other.

Two test call sites here still used the pre-#1099 three-argument `render_status`; they now pass `plain()` like the rest of the suite. Folded into the status commit rather than left as a fixup.

Verified on the rebased tree: `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace --locked --no-fail-fast` — clean apart from the three known `memory_import` `SUN_LEN` failures, which are a Unix-socket path-length limit from this worktree's long path and reproduce on an unmodified tree at the same path.

PR #1104, opened 2026-09-15
chore(ci): enforce that terminal colors stay behind the theme layer

BunsDev:
Closes the last unenforced invariant from #1084.

## Why

`docs/design/cast-tui-contract.md` §2.4 requires every accent color to come from a semantic token in `theme.rs`, and the Phase 2 done-when checklist states it outright:

> `rg "Rgb \{" crates/coven-cli/src/tui` returns nothing.

**Nothing enforced it.** Every contributor has complied by hand, but a single inline `Color::Rgb(0x9A, 0x8E, 0xCD)` would quietly fork the palette and drift from `brand/ui/color-tokens.css` — and `brand_tokens_mirror_color_tokens_css` would not catch it, because that test only validates tokens that go through `theme.rs` in the first place.

This matters more now than it did last week: #1099, #1100, #1098 and #1103 all add color paths, and #1100 adds a whole second (light) appearance.

## What it blocks

Outside `theme.rs`:

| Pattern | Why |
|---|---|
| `Rgb { .. }` | a hand-rolled brand token |
| `Color::Rgb(..)` / `RatColor::Rgb(..)` | a ratatui color built from channels instead of `theme::ratatui_color` |
| `38;2;` / `48;2;` | a hand-assembled SGR truecolor escape |

`theme.rs` is the one allowed file — it is the mirror of the canonical CSS and is already covered by the drift test.

## The test-module exemption

The escape rule is exempt inside `#[cfg(test)]` modules. Four existing assertions in `cast/render.rs` and `chat/highlight.rs` check that a rendered frame carries a token's escape:

```rust
// PRIMARY_STRONG (0x9A, 0x8E, 0xCD) for SAFE.
assert!(safe.contains("\x1b[38;2;154;142;205m"));
```

That is the invariant *working*, not breaking — my first version flagged all four, which is how I found it. A hand-rolled `Rgb {}` is still blocked everywhere including tests, since a forked palette is a forked palette wherever it lives.

## Status

Passes across all 243 Rust files today, so this pins the current state rather than asking for cleanup.

## Tests

`scripts/check-theme-tokens-test.py`, 10 cases: each blocked shape, the aliased `RatColor::Rgb` form, semantic-token usage passing cleanly, escape assertions inside a test module passing, a hand-assembled escape in production code still blocking, `Rgb {}` blocking even inside tests, and a case asserting the live tree satisfies the invariant.

Both the guard and its tests are wired into `ci.yml` beside the other policy checks; `check-ci-workflow-test.py` and `check-workflows.sh` both pass.


BunsDev:
The guard caught a false positive on its first real run, and the fix is in.

After #1099 merged, `Policy guard` went red on:

```
- crates/coven-cli/src/observe.rs:70: raw_ansi_truecolor
```

Line 70 is a **doc comment** — prose explaining the hazard that PR was guarding against:

```rust
/// escape in here reaches the user as literal `[38;2;154;142;205m` garbage.
```

So the guard was flagging documentation *about* escapes as if it were an escape. Ironic target, but a real defect in the check: comments are exactly where these patterns appear innocently.

`scan_text` now skips whole-line comments. Deliberately **only** whole-line ones — a trailing comment must not shield real code, so this is still caught:

```rust
let c = Color::Rgb(1, 2, 3); // brand violet
```

Two cases added: the `observe.rs` doc-comment shape verbatim (including `//!` and a commented-out `Rgb {}`) passes, and the trailing-comment line above still blocks. Suite is 12 cases, all green; the guard passes across all 243 Rust files on the rebased tree.

This is the second over-broad first draft I have had to narrow on this batch — the privacy-guard fix in #1097 had the same shape, where the obvious wide fix silently exempted sha512 digests. Worth noting that in both cases the existing suite, not review, is what caught it.

PR #1112, opened 2026-09-16
fix(afs): disable git hooks during materialization

BunsDev:
Clean rebase of #1095 onto current `main` (`a91fe147`). This PR carries only the vetted `crates/coven-cli/src/afs.rs` blob from the original aardvark fix, with current main as the direct parent.

Security intent: prevent repository-configured hooks from executing during AFS `git worktree add` / `git commit` materialization while preserving commit signing.

Original focused evidence: `cargo fmt --check`, package clippy, `afs::tests::commit_does_not_run_repository_configured_hooks`, secret/privacy checks, and diff checks passed. Current CI should revalidate the rebased commit.


BunsDev:
Closing this replacement rebase attempt. Validation showed the original aardvark head was based on an ancient merge-base, so transplanting the whole `afs.rs` blob onto current `main` would revert ~959 lines of later AFS work. This PR must not land. I’m rebuilding the security fix by applying only the original one-commit delta to current `main` instead.

PR #1094, opened 2026-09-15
fix: authenticate branch-triage PR heads and bind merges to gated commit

BunsDev:
### Motivation
- Bind the branch-triage merge to the exact commit the merge gate evaluated, so a push landing between gate evaluation and `gh pr merge` cannot ride an earlier green result onto `main`.
- Stop a merged **fork** PR whose `headRefName` collides with a live origin branch from getting that branch classified MERGED and deleted.

### Description
This branch was rebuilt from current `main` on top of #735's `triage.sh` rather than rebased; the original branch was 678 commits behind and conflicted in the script.

Kept (the two pieces still novel on `main`):
- **Merge binding.** Right after #735's trust check proves `origin/<branch>` is the commit GitHub reports as the PR head, record it as `gated_oid` and pass it to `gh pr merge --match-head-commit`. The merge API refuses if the head moved. (`--match-head-commit` needs gh >= 2.20; `ubuntu-latest` ships far newer.)
- **Same-owner filter on the merged-PR list.** `MRGD_PR_JSON` now requests `headRepositoryOwner` and keeps only PRs whose head lived under this repository's owner, via the same owner value `trusted_open_pr_filter` already compares against (hoisted into `REPO_OWNER` so both use one source). A branch that only matches a merged fork PR by name falls through to the unique-commit check instead of being deleted. Narrowing this list can only reduce deletions, so #735's deliberate fail-open, name-based protection of open PRs is untouched.
- README: documents the head check, the merge binding, why deletion stays name-based, and corrects the MERGED definition. Workflow: the `contents: write` comment no longer describes the rebase step that was removed.

Dropped as already fixed on `main`:
- Filtering `OPEN_PR_JSON` to same-repo heads and the pre-merge `headRefOid` / `origin/<branch>` comparison — #735 (c563f6b3) covers this with `trusted_open_pr_filter` at merge time, and intentionally keeps the open list name-based for deletion protection.
- `merge_state_json` / `merge_block_reason_from_json` split — `merge_block_reason` is kept as `main` has it.
- Nothing from the stale base is reintroduced: `main`'s `git for-each-ref` branch listing, `git fetch --prune -q`, and `refs/remotes/origin/` resolution (#624/#625/#630) stay as they are.

### Testing
- `bash -n .github/actions/branch-triage/triage.sh` — ok
- `shellcheck` — no findings on changed lines (two pre-existing SC2086 infos on untouched lines)
[...]

Issue #1051, opened 2026-09-13
Investigate native Windows IPC closure and pre-gate reservation latency failures


BunsDev:
## Distinct preserved observations

These are not asserted to share a cause and are not fixed by merely refreshing the PR base.

1. Protected-route head7681ebdf, CI34780123195: default Windows workspace failed `daemon::tests::discovers_same_profile_legacy_pipe_from_an_inherited_acl_status_file` with `failed to write Coven daemon request: The pipe is being closed. (os error232)`. CLI unit summary3201 passed/1failed/3ignored. Feature Windows job was green. Existing fake server handles health, metadata-only ACL connection, health in that order (`daemon.rs:13369`); failing phase was not retained precisely enough to conclude whether this is fixture accept ordering or transport behavior.
2. Corpus headf4d3a487, CI34780276230, Windows feature job103785946065: `output_auto_review_array_before` POST `/api/v1/familiars/sage/edits` received no response before the5s request deadline. E2E80passed/1failed; Linux/default Windows and all three feature companion targets passed. The fixture was set up, and last_rpc_observation records5,032,190us, owned child running, status present. Daemon checkpoints localize the delay: reservation-begin17,186us -> reservation-ready5,204,370us, then gate/identity proceed. Startup store setup took246ms, so this is NOT the prior13.436s startup observation. The phase encompasses reservation size accounting and canonical WardAuditReservation acquisition; it does not identify which operation was slow.

Artifact10325162622 has official SHA256 `5f9fb4619211041fe1696ec6a8b2c8cef4610b9c4fa7503c234366ee278471ff`. Failed run_id: `output-auto-review-shape-array-true-7504-4f48d1a6-17ee-47ab-91fd-22b672820352`. Actual CI Coven3795cd06ad5b7a20f17706f65bc72bf43ddc6ae3, committed core0021fd2662d0b82328371ee3a6957d645f776fda, no local override. Complete manifest/JUnit/logs retained.

## Investigation boundary

The fixture, daemon legacy test and client transport files are unchanged from base5d662749 through both candidate heads. This is not proof that no integration interaction exists. Preserve failures, trace actual connection/request boundaries, and establish a discriminating reproduction before remediation. Do not retry writes into success, weaken fail-closed behavior, widen documented budgets, or describe green later CI as a causal repair.

[...]

Issue #1031, opened 2026-09-12
Resolve slow owned Windows startup exposed by final boundary acceptance


BunsDev:
Follow-up to #1030, tracked under Threads threads-vpp / threads-zav. Native run34696507398 at a8486f407d518398f5e6ac1c4866101e0d0d86d1 (df0ccd91 source tree) passes all six original dedicated CLI lifecycle cases and five new diagnostic cases. The CLI repair is not the new failure. Feature threads_e2e is69/70: clock_control_and_audit_time_survive_real_daemon_restart times out during initial owned serve after15.0065678s, before setup. PID4052 was terminated by fixture cleanup; its exit1 is not a natural daemon failure. Recovery checkpoints show store-initialize-end at13436ms, after connection816 / Ward2907 / runtime3626 / schema3634 / commit8136. No daemon-store-end or status exists before termination. Artifact10298638899 SHA2560596f02019af72dfb75bfd59aaeb9430e1372ef1a1ceb93a4e8fa95d18162164 retained. Investigate the production store path and cold IO; do not invent missing phase timings, relax auth/durability, extend budgets without causality, or count the next docs-only rerun as a fix. Scope belongs to existing #1029 contribution, not a competing integration. Acceptance: evidence-driven correction, focused regression, fresh native and actual current-core boundary proof, preserved failed receipt. Parent root issues and human gates remain open.

BunsDev:
Production correction published in #1029 at bec643fbbde7857621f73dd75d01719aa6138f3b. Deterministic regressions reproduced an unnecessary initialize/commit -> last-connection checkpoint -> reopen -> hub/cache -> second close in both production store-open paths. The initialized connection now receives the same runtime guards, remains open through setup, and explicitly closes once BEFORE readiness. No migration SQL, synchronization pragma, authority, deadline or dependency changes; no lifelong keeper or shifted shutdown checkpoint. WAL retention, audit guard, autocommit/unpinned checkpoint, normal cleanup and process-exit recovery regressions pass. Completed-prior-append accounting now separates observed logger cost without inventing the current append duration. The native 5300ms interval is not proven pure SQLite close time or time saved. Parent reviewed the source and reran4newunits+5diagnostics; wider local boundaries and scoped cross checks pass. Fresh native/current-checkout acceptance is starting. The docs-only 34697650776 pass is retained as same-runtime comparison, not causal repair.

BunsDev:
[...]

PR #1132, opened 2026-09-18
fix(store): stop fsyncing every commit and checkpointing on every request

BunsDev:
## Context

- Summary: `Threads daemon tests (Windows)` fails on `main` and in the merge queue with `no response bytes arrived before the deadline`, a different test each time, always a scheduled-submission `POST`. This removes the per-request fsync and checkpoint cost that the retained diagnostics measured, so the request finishes inside the client's 5 s budget without touching any deadline, retry, or budget.
- Files changed: `crates/coven-cli/src/store/schema.rs`, `crates/coven-cli/src/store.rs`, `crates/coven-cli/src/threads_request_diagnostics.rs`, `crates/coven-cli/tests/support/threads_output_auto_cases.rs`
- Refs #1051 (item 2, the runtime latency observations; item 1 closed via #1082). On 2026-09-18 this flake evicted the merge queue roughly every other batch (#1094, #1124 ×3, #1086, #1092).

## Implementation

- Approach: read the request as a sequence of disk syncs. Each scheduled-submission POST ran on one thread: `open_store` (rebuilds the TEMP capacity trigger and table on every open), `WardAuditReservation::acquire` (store-identity `INSERT OR IGNORE`, an unconditional `wal_checkpoint(PASSIVE)`, `BEGIN IMMEDIATE`/`COMMIT`), the gate work (`publish_new` with `sync_all`, the audit `COMMIT`), then `finish` (`BEGIN IMMEDIATE`/`COMMIT`, another `wal_checkpoint(PASSIVE)`). With SQLite's default `synchronous=FULL` every `COMMIT` fsyncs the WAL and every checkpoint fsyncs the database. The retained packets on #1051 put the time exactly there: store-open→ready 1.3–1.5 s, acquire-side PASSIVE 1.8–2.6 s, reservation COMMIT 1.4 s, release-side PASSIVE ~1 s, 6–7 s total. The 30 s proposal scheduler, 60 s maintenance pass, and 60 s automations tick checkpoint on their own cadence and serialize with the request on the checkpoint lock, which is why the failing test is random.
- Changes:
  1. Writable connections set `PRAGMA synchronous = NORMAL` (the documented pairing for WAL: durable across a daemon crash, WAL synced before each checkpoint, no fsync per commit; only an OS crash or power loss can lose the final commits) and `PRAGMA temp_store = MEMORY` (the per-open TEMP objects no longer create a file under the OS temp directory).
  2. `acquire` checkpoints only when the admission would otherwise fail the WAL budget, using `TRUNCATE` as before; the unconditional PASSIVE pass is gone. `finish` no longer checkpoints. `wal_autocheckpoint` (4 MiB) and the maintenance pass bound the WAL otherwise, and a pinned WAL still fails closed with 507.
  3. `get_or_insert_store_meta` reads before it writes, so the hot-path identity bind no longer takes the write lock when the key exists.
- User-visible behavior: none at the API. The `reservation-passive-begin/ready` and `reservation-release-passive-ready` diagnostic phases no longer exist (they described the removed work); `reservation-truncate-*` still appears when a reclaim happens.
- Compatibility notes: the `synchronous` assertion in `open_store_retains_initialization_wal_without_close_and_reopen` changes from `>= 2` to `== 1` with the rationale in the comment; #1031's commit said "do not alter durability pragmas" as a scoping statement for that fix, and no documented durability contract references fsync-per-commit (grepped `docs/`, `specs/`, `SECURITY.md`).

## Verification

- [x] `cargo fmt --check`
- [x] `cargo clippy --workspace --all-targets --features threads-test-clock -- -D warnings`
- [x] `cargo test -p coven-cli --bin coven --features threads-test-clock store::tests` — 188 passed
- [x] `cargo test -p coven-cli --bin coven --features threads-test-clock ward_audit` — 44 passed, including `ward_audit_wal_checkpoint_reclaims_idle_wal_but_rejects_pinned_wal` unchanged
- [x] `cargo test -p coven-cli --bin coven --features threads-test-clock request_diagnostics` — 7 passed
- [x] Real-daemon Threads suites on macOS, `--no-fail-fast`: `threads_e2e` 104 passed, `threads_identity_invariants` 45 passed, `threads_protected_intake` 39 passed, `threads_terminal_recovery` 34 passed (0 failed, 0 ignored). Includes every case that failed on Windows in the last two days.
- [x] `check-coven-privacy.py --range` and `check-docs-ownership.py --range` on `origin/main...HEAD`
- [x] Policy guard loop from AGENTS.md — `Policy guard (local): 0`
- [ ] Additional manual checks: native Windows acceptance is CI's; the claim this PR makes is that the measured intervals were sync cost, and the Windows Threads job over the next runs is the evidence.

## Risk and Rollback

- Risk level: medium. `synchronous=NORMAL` narrows the durability guarantee from "every commit survives power loss" to "every commit survives a daemon crash"; the store is per-user local session state and the daemon is the only writer. Skipping the release-side checkpoint means the WAL can grow between admissions up to the autocheckpoint threshold instead of being reclaimed after every write.
- Rollback plan: revert the commit. No schema or on-disk format change; `synchronous` and `temp_store` are per-connection.

## Agent Handoff

- Current state: complete.
- Follow-ups: if the Windows Threads job still exceeds 5 s after this lands, the remaining measured interval to chase is store-open→store-ready (the TEMP trigger rebuild per open); caching the configured connection per request thread would remove it.
- Known gaps: no native Windows run before opening; the phase-order assertion in `threads_output_auto_cases.rs` was updated to match the removed phases.


BunsDev:
Native Windows evidence from this PR's own `Threads daemon tests (Windows)` run (35389640822, passed) against the most recent failed `main` run (35309646993), taken from the retained `threads_request_checkpoint` records across every test home in each job's artifact:

| phase (elapsed from request start) | main 35309646993, failed | this PR 35389640822 |
| --- | --- | --- |
| `store-ready` | n=122, mean 16 ms, max 362 ms | n=123, mean 7 ms, max 62 ms |
| `reservation-ready` | n=122, mean 119 ms, max 721 ms | n=123, mean 9 ms, max 64 ms |
| `response-begin` (all POSTs) | n=289, mean 209 ms, max 3363 ms | n=295, mean 100 ms, max 3055 ms |

The reservation path, which carried the two checkpoints and the fsync'd commits, went from a 119 ms mean with a 721 ms tail to 9 ms with a 64 ms tail. Mean response time halved.

The one remaining multi-second sample on the PR run is a single `stage-ready` at +3035 ms: `proposal_store::publish_new`'s `sync_all` on the staged proposal file, one `FlushFileBuffers` that the runner served slowly once in 295 requests. It stays under the 5 s budget on its own now that nothing else on the path syncs, and it is the deliberate durability point for the staged file, so it is left alone here and noted as the next interval to look at if the job ever exceeds budget again. On the failed main run the slowest request was a tick whose scheduler cursor pass spent 3.3 s, with `reservation-passive-ready` and `reservation-commit-ready` at 204 ms and 221 ms on top.

